PR #29, opened 2018-10-29
Fix/confusing space and tabs

corryvrequan:
There have been confusing tabs and spaces in codes.
Now all the tabs are replaced to four-spaces.


corryvrequan:
Note:
As far as all the indents are spaces or tabs, there is no problem.
However, there are many lines such that "the first indent is tab and second indent is four spaces". 
Therefore, we decided to use four-spaces as a indent.

PR #30, opened 2018-10-29
Fix/gcc8 issue

corryvrequan:
Fix errors caused by gcc-8, pybind11, and complex libraries.

### bug info 
When GCC version >= 8, pybind11 uses -std=c++14. 
However, we cannot use GCC builtin complex literals in this mode.
Therefore, cppwrap for python cannot include files using complex literals

### Fix
Force using gnu++11 when we have gcc-8.
Add 
set(PYBIND11_CPP_STARNDARD -std=gnu++11)
to CMakeLists.txt


TenninYan:
Don't send pull request to master branch!

Issue #33, opened 2018-10-30
Python crashed after calling `Observable::add_operator(PauliOperator* mpt)`


hal0taso:
# Expected Behavior
In python, add PauliOperator to Observable successfully (with no crash).

# Current Behavior
after calling `qulacs.observable.add_operator(mpt)`, python crashed.

# how to Reproduce
run following python script
```python
import qulacs

obs = qulacs.Observable(2)
pauli = qulacs.PauliOperator("Z 0 X 1",  1.0)
obs.add_operator(pauli)
```

PR #34, opened 2018-10-30
bugfix for qulacs.Observable.add_operator(PauliOperator*)

hal0taso:
# Description
Fix python crashing after calling `qulacs.observable.add_operator(mpt)`.

# Related Issue
#33 
I think this is caused by double-free.

# Update
- make argument to const
- create copy object from `mpt` and add to `_operator_list`


TenninYan:
元のコードを走らせた時にtestがfailするコード書けますか？

PR #35, opened 2018-10-30
Fix bug in Installation

yurisan:
#26 

### Update
- Use `VERSION_LESS`, instead of `VERSION_GREATER_EQUAL`(It doesn't work on cmake==3.5.1)
- Add `-dumpfullversion` flag to getting version of `gcc-7`, `g++-7`


corryvrequan:
LGTM

PR #37, opened 2018-11-09
fix bugs in transition amplitude, and add tests for them

corryvrequan:
Numerical bugs in transition amplitude are fixed.
Test functions for transition amplitude are also added.


hal0taso:
LGTM 👍

PR #45, opened 2018-11-19
[bugfix] fix double-free and invalid depth-evaluation in probabilistic gate

corryvrequan:
- Bug description
When we call QuantumCircuit::to_string with a quantum circuit with ProbabilisticGate in python, the kernel crashes.

- Reason
Since the number of target qubits in ProbabilisticGate cannot be determined before execution, it returns zero as a temporal value. 
However, a function QuantumCircuit::to_string crashes when it encounters any gate of which the number of target qubits is zero.

- Update
The behavior of QuantumCircuit::to_string is changed, and it now ignores any gate with empty target qubit list.
I also added small test function for checking to_string with probabilistic gate.


corryvrequan:
- Bug description
When we add ProbabilisticGate generated by functions such as DepolarizingNoise to a circuit in python, and a circuit accesses the gate, python kernel crashes.

- Reason
When we add probabilistic gate to a python in this way, python releases instances which are still used in C++. This cause that circuit accesses released gate and use it for state updates.

- Update
We redirect circuit::add_gate in python to circuit::add_gate_copy in C++,
and make circuit::add_gate_consume in python which is redirected to circuit::add_gate in C++.

This makes C++-code independent of python-related pointer as default behavior.
Main drawbacks of these changes are that "add_gate" slows down since it copies gate matrix, and it requires additional memory for copying gate matrix.

Issue #67, opened 2018-12-10
[Bug] ParametricQuantumCircuit.copy() returns a instance of QuantumCircuit


kosukemtr:
Not in hurry, but someone please fix this.

corryvrequan:
This problem is solved in PR #71 . Thanks!

PR #71, opened 2018-12-10
Implement many features

corryvrequan:
# Update

## [Feature] Change the factor of the rotation gates.

I have changed the factor of the rotation angles of RX, RY, RZ, PauliRotation, and these parametric version from 1.0 to 0.5.
Now unitary matrix of RX(\theta) is exp(i\theta/2 X). This was exp(i\theta X).
I also have fixed test functions so as to be compatible with these changes.

**This changes break backward compatibility of numerical results.**

## [Bugfix] Fix copy function  for ParametricQuantumCircuit class. 

This bug is reported in issue #67 by @kosukemtr .

ParametricQuantumCircuit::copy() returns an instance of QuantumCircuit without copying information of parametric gates.
I've fixed this bug, and now it returns an instance of parametric quantum circuit.

```python
from qulacs import ParametricQuantumCircuit
qc = ParametricQuantumCircuit(3)
qc.add_parametric_RX_gate(0,0.1)
qc_copy = qc.copy()
print(type(qc_copy))
print(qc_copy.get_parameter(0))
print(qc_copy)
```
This code outputs
```
<class 'qulacs.ParametricQuantumCircuit'>
0.1
*** Quantum Circuit Info ***
# of qubit: 3
# of step : 1
# of gate : 1
# of 1 qubit gate: 1
Clifford  : no
Gaussian  : no

*** Parameter Info ***
# of parameter: 1
```

## [Feature] Add many error messages for invalid arguments.

In the previous version, if invalid arguments are given to functions such as QuantumCircuit::get_gate(index), segfault happens and python-kernel will crash.
I've added many error messages for invalid arguments as follows.

- Try to get out-of-index gate, parameter, or pauli_operator, 
- Try to add/remove quantum gate to circuit which applies to invalid qubit index (e.g. circuit.add_gate(X(3)) for 2-qubit circuit.)
- Try to add Pauli operator to observable which applies to invalid qubit index (e.g. observable.add_term(0.1, "X 3") for 2-qubit circuit.)
- Try to apply quantum circuit or observable to quantum state with incompatible size.
- Try to access invalid index of quantum state (e.g. measure n+1-th qubit in n-qubit state)
- Try to make 5-th single-qubit Pauli operator.

For example, the following code will not exit with error, 
```python
from qulacs import QuantumCircuit
qc = QuantumCircuit(3)
qc.add_X_gate(100)
```
but output the following error message to error stream.
```
Error: QuatnumCircuit::add_gate(QuantumGateBase*): gate must be applied to qubits of which the indices are smaller than qubit_count
```
The same message is shown in C++ program.
Note that I've checked all the test function outputs no error messages.
I hope this is helpful for solving segfault.

## [Feature] Change test functions for Hamiltonian decomposition

In the case when we cannot find files of openfermion in the test of Hamiltonians, the program will exit and the following test functions are not done.
I will remove std::exit in assertion, and add assertion macros.

### DIFF
--- a/doc/cppsim/3_Tutorial_python.md
+++ b/doc/cppsim/3_Tutorial_python.md
@@ -214,6 +214,9 @@ del dense_gate
 - Measurement : Measurement
 - Noise : BitFlipNoise, DephasingNoise, IndepenedentXZNoise, DepolarizingNoise
 
+回転ゲートである<code>RX</code>,<code>RY</code>,<code>RZ</code>,<code>PauliRotation</code>は所定のパウリ演算子$P$について、引数$\theta$に対して$\exp(i\frac{\theta}{2}P)$という操作を行います。
+それぞれのゲートの詳細はAPIドキュメントを参照してください。
+
 
 ### 量子ゲートの合成
 続けて作用する量子ゲートを合成し、新たな単一の量子ゲートを生成できます。これにより量子状態へのアクセスを減らせます。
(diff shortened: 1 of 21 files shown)


corryvrequan:
I would like to ask review the first three updates for @kosukemtr , and the last two updates for @hal0taso .

hal0taso:
Since this value also be able to be passed as arguments, and considering trade-off of precision and runtime, the problem will be not big even if we keep it as it is.

PR #86, opened 2019-01-02
Make gpu files compatible with linux

corryvrequan:
GPU related files were only tested in MSVC.
I've checked them on linux, and fix some codes for compatibility.


corryvrequan:
CPU-related codes are untouched.
This PR is self-merged.

PR #91, opened 2019-01-11
Fix error in optimization for quantum gates with unordered qubit indices

corryvrequan:
- Bug
When we use quantum circuit optimizer, gate will be greedily merged as far as the size of merged gate does not exceed  given <code>block_size</code>. To this end, circuit optimizer has a function to compute the size of merged gates before merging. However, this function does not output correct answer for two gates of which the <code>target_qubit_index_list</code> is not sorted.
This potentially causes insufficient or over-optimization according to <code>block_size</code>.

- Update
I've fixed codes in gate optimization, and added test with unsorted gates.
Note that I think numerical results do no change with this bug and update.


corryvrequan:
Since this is small bugfix, this PR is self-merged.

PR #100, opened 2019-02-01
Add retrying for the Chocolatey commands in Windows CI

### LINKED ISSUE #48: Failing CI in Windows when the chocolatey server is down
- https://travis-ci.org/qulacs/qulacs/jobs/457537327#L70

### Methods to solve this issue

1. Ignore the result of Windows CI
    - We can use `allow_failures` to ignore the result of CI in Windows
        - See also: https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail
    - But it makes CI meaning to be unclear....
2. Implement retry and backoff logic to the chocolatey server in `.travis.yml`

y-yu:
Fix #48 .

Travis CI has a retry function: [`travis_retry`](https://docs.travis-ci.com/user/common-build-problems/#travis_retry) so I try this to prevent us from failing Windows CI due to Chocolatey server down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,8 +103,8 @@ cache:
 before_install:
   - |
     if [[ "${TRAVIS_OS_NAME}" == "windows" ]]; then
-      choco install -y cmake
-      choco install -y python --version "${PYTHON}"
+      travis_retry choco install -y cmake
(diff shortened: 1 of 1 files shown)


corryvrequan:
LGTM.
Thank you for nice PR!

PR #101, opened 2019-02-04
[bugfix] Change return value policy for avoiding memory leak

corryvrequan:
- Update

Change return value policy of copy function and gate factory from automatic_reference to take_ownership.

For checking there is no segfault in typical codes, I have added some python tests on pointer handling.

- Reasons

When we use function such as <code>qulacs.gate.X(index)</code>, a new gate instance is created in C++ side, and python refers its pointer without ownership. Since this pointer is not released in C++, this causes memory leak. In particular, this problem is serious when we create a lot of heavy QuantumGate object because it eventually leads to out-of-memory error.

- Solution

By changing return value policy from automatic_reference to take_ownership, python-side is responsible for releasing given pointers.

- Note

QuantumCircuit is responsible for releasing added gate instances. To avoid double-free, <code>QuantumCircuit::add_gate</code> function called from python uses copied gate instance of arguments as default. Therefore, QuantumCircuit is (I hope) valid even if gate is released after <code>QuantumCircuit::add_gate</code>.

Though I carefully examined return value policies, I cannot cover all the possible pointer handling codes and there may be segfault scenarios. Reports on potentially-segfault-code are welcome.

### DIFF
--- a/python/cppsim_wrapper.cpp
+++ b/python/cppsim_wrapper.cpp
@@ -62,7 +62,10 @@ PYBIND11_MODULE(qulacs, m) {
         .def("get_qubit_count", &GeneralQuantumOperator::get_qubit_count)
         .def("get_state_dim", &GeneralQuantumOperator::get_state_dim)
         .def("get_term_count", &GeneralQuantumOperator::get_term_count)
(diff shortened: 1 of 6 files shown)


corryvrequan:
This PR is approved by @kosukemtr .

PR #104, opened 2019-02-06
[feature] Reversible boolean function gate

corryvrequan:
# Overview
I've added new quantum gate <code>ReversibleBooleanGate</code>.
This gate supplies easy way to implement classical reversible function as an quantum gate.
This gate requires <code>target_qubit_index_list</code> and <code>function_ptr</code> as arguments. <code>function_ptr</code> is a pointer to a function which maps <code>(uint64, uint64) -> uint64</code>. This gate moves a value in <code>state[index]</code> to <code>state[func_ptr(index,gate_dim)]</code>, where <code>gate_dim</code> is <code>1<<target_qubit_index_list.size()</code>.

Thanks to pybind11, we can easily send python function to cpp-side, while its performance may be worse than pure C++ codes.

This code would be useful for the following purpose.

- Implement raising/lowering operator or generalized Pauli operator in multi-level systems.
- Implement integer function for number theoretic algorithm such as Shor's factoring.

This code is checked in MSVC 2017 on Windows 10 and gcc7.0 on Ubuntu 18.04. 
I've added small tests for this gate.

## Example: python

```python
from qulacs.gate import ReversibleBoolean

def func(index, dim):
    return (index+1)%dim

state = QuantumState(3)
print(state)
gate = ReversibleBoolean([2,0],func)
for _ in range(4):
	print(state)
```
With this code, state changes |000> -> |100> -> |001> -> |101> -> |000>. (Note that the rightmost qubit is the 0-th index.)
If you use <code>ReversibleBoolean([0,1],func)</code>, state changes |000> -> |001> -> |010> -> |011> -> |000>.

## Example: C++

```cpp
std::function<ITYPE(ITYPE,ITYPE)> func = [](ITYPE index, ITYPE dim) -> ITYPE {
	return (index + 1) % dim;
};
std::vector<UINT> target_qubit = { 2,0 };
auto gate = gate::ReversibleBoolean(target_qubit, func);
gate->update_quantum_state(&state);
```
Its behavior is the same as python code.


corryvrequan:
Since this PR does not affect other codes probably, this is self-merged.

PR #108, opened 2019-02-15
[Feature] Implementation of state reflection gate and sparse matrix gate

corryvrequan:
- Update

I've added two gates, StateReflection and SparseMatrix.

<code>StateReflection</code> gate is characterized by quantum state |s><s|, and its unitary matrix is given by U = 2|s><s| - I. This gate is frequently used in algorithms such as Grover's search or HHL.

<code>SparseMatrix</code> gate is similar to DenseMatrix gate but it holds its matrix elements as sparse matrix. Simulation of this gate is much faster than DenseMatrix when a gate acts on many qubits and its gate matrix is sparse.

These features are proposed by @Yuya-O-Nakagawa .

- Usage

Reflection Gate
```python
from qulacs.gate import StateReflection

n = 3
state = QuantumState(n)
reflect = QuantumState(n)
gate = StateReflection(reflect)
# Note: QuantumState object is copied when gate instance is created. 
# Since it might consumes a lot of memory, delete reflection state after creating gate if you use large quantum state.
del reflect
```

SparseMatrix
```python
from qulacs.gate import SparseMatrix
import numpy as np
from scipy.sparse import lil_matrix

n = 3
state = QuantumState(n)

# Note: Unlike np.ndarray, real sparse matrix is not implicitly converted to complex sparse matrix. 
# Thus, we need to explicitly define complex sparse matrix. 
matrix = lil_matrix( (2**n, 2**n), dtype = np.complex128 )
matrix[0,1] = 1
matrix[1,0] = 1
gate = SparseMatrix([0,1], matrix)
print(state)
```

### DIFF
--- a/python/cppsim_wrapper.cpp
+++ b/python/cppsim_wrapper.cpp
@@ -239,11 +239,13 @@ PYBIND11_MODULE(qulacs, m) {
     QuantumGateMatrix*(*ptr2)(std::vector<unsigned int>, ComplexMatrix) = &gate::DenseMatrix;
     mgate.def("DenseMatrix", ptr1, pybind11::return_value_policy::take_ownership);
     mgate.def("DenseMatrix", ptr2, pybind11::return_value_policy::take_ownership);
+	mgate.def("SparseMatrix", &gate::SparseMatrix, pybind11::return_value_policy::take_ownership);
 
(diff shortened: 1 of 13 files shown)


corryvrequan:
Since this PR does not affect other functions, it is self-merged.

PR #112, opened 2019-02-27
call specific functions in multi_qubit_dense_matrix_gate_host when the number of target qubit is small

puyokw:
When the number of target qubit is 1, 2, 3, and 4, we have changed to use specific functions in multi_qubit_dense_matrix_gate. 
Add a function for more than 11 qubits to multi_qubit_dense_matrix_gate.


corryvrequan:
Thank you for fixing and adding codes.
As far as I read, codes seem correct. Since I cannot verify GPU code for a few days, I will merge this PR, and open issue if I cannot run it in my environment.

PR #114, opened 2019-03-06
[bugfix] fix the bug of not called the gpu codes when calculating expectation value

puyokw:
fix the bug of not called the gpu codes when calculating expectation value


corryvrequan:
LGTM. Thanks for bugfix!

PR #115, opened 2019-03-08
calculating expectation value faster with gpusim

puyokw:
- Tuning parameters to calculate expectation value faster with gpu.
- Add some test cases to gpusim


corryvrequan:
Thank you for PR. LGTM.
(Probably, we can automatically generate gpu-test code from cpp ones since they are almost the same...)

PR #120, opened 2019-04-26
bugfix for gpu

corryvrequan:
Some bugfix for GPU codes.

1. gate_matrix_sparse.cpp: fix typo. std::end -> std::endl
2. gate_reflect.hpp: fix wrong line.
3. CMakeLists.txt in gpusim: CUDA_ADD_CUBLAS_TO_TARGET seems outdated in the latest cmake. I've added "target_link_libraries(gpusim_static/shared -lcurand -lcublas)" instead.

I've confirmed all the tests pass in current google colaboratory, but it needs check with other environments.


corryvrequan:
Note:
Current versions in google colaboratory are
CUDA: V10.0.130
gcc: 7.3.0
cmake: 3.12.0

corryvrequan:
With small fixes, I've checked it works in windows.
Environment is 
MSVC: Visual Studio 14 2015 Win64
CUDA: 9.0.176
cmake: 3.12.0-rc2

corryvrequan:
This is also verified on AWS P2 instance with GPU K80 and
Ubuntu 16.04
CUDA9.2 + cuDNN 7
gcc 7.3.0
python3.6

PR #121, opened 2019-04-29
Update gcc, macOS and Python versions

y-yu:
- I try to update the version of gcc and g++ to 8 from 7
    - I fix some versions of `CMakefile.txt` and so on
- Python2.7.11 requires the version of OpenSSL less than 1.0
    - But `homebrew` does not provide OpenSSL whose version < 1.0 then I decide to update Python2.7.11 to 2.7.15 (latest)
        - That's the why #197 and #196 failed in those CI
    - In Windows we cannot help but to keep using 2.7.11 since `chocolatey` does not provide 2.7.15
- Likewise Python3.6 also requires version of OpenSSL less than 1.0 so I update that version into 3.7.5(latest) too
- Since updating OpenSSL, Ubuntu 14.04(tursty) does not provide OpenSSL1.0 so I have to update Ubuntu version to 18.04(bionic, latest LTS)
    - At this distribution CMake3 is default version so I don't need to specify `cmake3`
- macOS version up is not necessary but there is no reason why we're using old version
- If there is `pyenv` in the Travis CI cache, but there may be not the Python interpreter. I add some code to install Python when it does not exist
- I add `using namespace std::complex_literals;` to take some compile time errors down. I don't know if it's right or not.....


y-yu:
I give up Python's version up for now....

corryvrequan:
I've checked commits. Thanks for your contribution!

PR #130, opened 2019-05-22
Add cuda stream version

puyokw:
- debug gpu version of transition_amplitude_multi_qubit_pauli_operator_Z_mask
- add functions of cudaStream version in order to launch several kernels simultaneously
- add tests


corryvrequan:
I've tested this PR works also in my environment.
Thanks for nice PR!

PR #132, opened 2019-06-11
setup.py picks up gcc / g++ command from environment variables

ssmi1975:
This may conflict with the effort in #121 , but this is a PR for setup.py to use gcc / g++ specified by environment variables.

With this, you can specify the compiler as follows,

```shell
CC=gcc-9 CXX=g++-9 python setup.py install
```

I haven't tried yet, but this should also work when installed through pip if it's ready in PyPI.

```shell
CC=gcc-9 CXX=g++-9 pip install qulacs
```

Because you can now specify the gcc / g++ command explicitly, setup.py will fail if the detected version is smaller than `7.0.0`. Please remove it if it is too much.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
  
 # Qulacs
 [![Build Status](https://travis-ci.org/qulacs/qulacs.svg?branch=master)](https://travis-ci.org/qulacs/qulacs)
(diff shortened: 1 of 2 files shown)


ssmi1975:
I'm afraid I have branched off from a wrong one. closing

PR #147, opened 2019-07-12
Bugfix of two qubit gate in csim

corryvrequan:
There was a bug in <code>two_qubit_dense_matrix_gate</code>.
I've fixed this bug and add test for this function.

Note that this function is currently not used from cppsim. Thus this is not tested and does not affect the behavior of C++/python libraries. 
This bug is reported in issue #145 by @k-ishizaka. Thanks for bug reporting!


corryvrequan:
There seems no issues in this PR. I'll merge this PR.

PR #148, opened 2019-07-12
Two qubit depolarizing gate

corryvrequan:
I've added <code>TwoQubitDepolarizingNoise</code>.
This applies two qubit pauli operators except identity to given two qubits with probability <code>prob/15</code> for each. This is equivalent to convert density matrix as rho -> (1-p)rho + p \* rho, where p = prob \* 15/16.

This feature is suggested in issue #146 .


corryvrequan:
There seems no issues in this PR. I'll merge this PR.

PR #149, opened 2019-07-14
Fix cmake script for GPU on ubuntu

shukob:
Hi, it did not build without this modification. Would you please check?


corryvrequan:
Thanks for PR, @shukob ! And, we are very sorry for the inconvenience in GPU build.
Though we checked it works in our environment, such as google colab and AWS GPU instance, we know our GPU build scripts still fail sometimes. So, this PR is highly appreciated.
As far as I checked variable names in cmake documents, the modified one seems better way to specify libraries. We will check that it works in our existing test envs, and then merge it. 

During checking this PR in our envs, can I ask the followings about this  PR, if possible?

1. Please change target of this PR from master branch to dev branch. Basically we modify master only in version update.
2. Please list versions of your (ubuntu, gcc, cmake, CUDA). We wanna manage list where we can successfully build.

shukob:
@corryvrequan Thank you for the comments. Re-requested on https://github.com/qulacs/qulacs/pull/150

PR #152, opened 2019-07-31
Parametric gate does not work with density matrix 

corryvrequan:
Parametric quantum gates such as <code>ParametricRX</code> cannot update recently added <code>DensityMatrix</code>. I've added update codes and now it works. I also added some tests about this issue.
This bug is reported in slack community. Thanks!


corryvrequan:
Since this is a bug patch about an important feature, I'll merge it with my test.
If there is a problem, please open issues.

Issue #151, opened 2019-07-31
build failure on Ubuntu 18.04 (CPU only)


miyo:
Build failure occurred on my environment, Ubuntu 18.04 with gcc 7.4.0 and cmake 3.10.2.
```
$ git clone https://github.com/qulacs/qulacs.git
$ python3 setup.py install --user
[ 54%] Building CXX object src/cppsim/CMakeFiles/cppsim_static.dir/circuit.cpp.o
In file included from /home/miyo/src/qulacs/src/cppsim/circuit_optimizer.hpp:4:0,
                 from /home/miyo/src/qulacs/src/cppsim/circuit_optimizer.cpp:4:
/home/miyo/src/qulacs/src/cppsim/type.hpp:5:10: fatal error: csim/type.h: No such file or directory
 #include <csim/type.h>
          ^~~~~~~~~~~~~
compilation terminated.
src/cppsim/CMakeFiles/cppsim_static.dir/build.make:86: recipe for target 
...
subprocess.CalledProcessError: Command '['cmake', '--build', '.', '--target', 'python', '--config', 'Release', '--', '-j2']' returned non-zero exit status 2.
```
It seems that there does not exist include files in expected directories. After the following quick hacks, the build is succeeded.
```
 mkdir include/{csim,cppsim,vqcsim}
 ln -s `pwd`/src/csim/*.{h,hpp} `pwd`/include/csim
 ln -s `pwd`/src/cppsim/*.hpp `pwd`/include/cppsim
 ln -s `pwd`/src/vqcsim/*.hpp `pwd`/include/vqcsim
```
Could you tell me your correct build way?

corryvrequan:
Thanks for reporting problems. I've tested it in a similar situation, but I could successfully build couldn't regenerate the above errors.
Thus, I wanna ask more detail about this error if possible. 

In a correct senario, cmake will copy all the (*.hpp and *.h files) in (csim,cppsim,vqcsim) to 
a folder "./qulacs/include/ folder". If you don't have "include" folder or don't have any header files or folders in "include" folder, there is a something wrong in the beginning of cmake build.

Can I ask followings for solving the problem more?
1. Can I ask a detailed situation about "include" folder?
2. Is there any warning/error message which seems to be related to this issue at the beginning of cmake build?
3. Can you try "./script/build_gcc.sh" command from qulacs folder? This will build qulacs without python setuptools.

Note that I used the following environment.
[...]

PR #155, opened 2019-08-01
Fix a problem in CMake script which didn't work when path contains "./src/"

corryvrequan:
When qulacs is installed cmake, cmake will copy all the *.hpp and *.h files in ./qulacs/src/*/*.(hpp|h) to ./qulacs/include/*/*.(hpp|h) by replacing "/src/" to "/include/". This does not work when qulacs folder itself contains "/src/" in its full-path. This problem would happen not only in source build but also when we try to install qulacs with pip and python path contains "/src/". Now cmake build script is modified so that only replace src in source directory. 
This issue and patch is made by @miyo in issue #151 . Thanks for contribution!


y-yu:
LGTM

PR #159, opened 2019-08-28
Show error message, return NULL ptr, and raise exception for invalid gate instantiation

corryvrequan:
# Overview
Add features to show message when gates act on the qubits with the same indices.
This feature is proposed at issue #158 by @kenjikun. Thanks for proposal!

# Detail
In quantum gates, we cannot mathematically define CNOT with same control and target, or unitary acts on the same indices twice.
```cpp
auto cnot = gate::CNOT(0,0);
auto random_unitary = gate::RandomUnitary({0,1,2,1})
```
In the previous implementation, however, the above codes will not check duplication and will gate-apply will work almost all the cases. Since this bug will not show segfault or nan/inf, once this kind of bug happens, it is hard to debug this.
With this PR, qulacs will show message in stderr and return null pointer. When you try to apply gate to state or add gate to circuit, null pointer segfault will happen.

In python case,
```cpp
cnot = qulacs.gate.CNOT(0,0);
random_unitary = qulacs.gate.RandomUnitary({0,1,2,1})
```
will raise ValueError exception through pybind11.

In the case of python, qulacs raises exception for the following cases.
- (multi-qubit gate) Qubit indices where gate acts on are duplicated.
- (matrix gate) Matrix dimension is not 2^(#qubit).
- (Pauli gate) Qubit count and Pauli-id count is not equal.

# Note
This is a common error since we tend to use the following codes when we try to choose indices randomly for m-qubit gate for n-qubit circuit.
```python
targets = np.random.randint(n ,size=m)
```
For removing duplication, one possible single-line code is 
```python
targets = np.random.permutation(np.arange(n))[:m]
```

### DIFF
--- a/python/cppsim_wrapper.cpp
+++ b/python/cppsim_wrapper.cpp
@@ -279,30 +279,83 @@ PYBIND11_MODULE(qulacs, m) {
     mgate.def("RY", &gate::RY, pybind11::return_value_policy::take_ownership);
     mgate.def("RZ", &gate::RZ, pybind11::return_value_policy::take_ownership);
 
-    mgate.def("CNOT", &gate::CNOT, pybind11::return_value_policy::take_ownership);
-    mgate.def("CZ", &gate::CZ, pybind11::return_value_policy::take_ownership);
-    mgate.def("SWAP", &gate::SWAP, pybind11::return_value_policy::take_ownership);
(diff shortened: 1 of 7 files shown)


corryvrequan:
There seems no error in this PR, this is self-merged.

PR #160, opened 2019-09-05
add gpu version of inner_product to python

puyokw:
add gpu version of inner_product to python


corryvrequan:
I've verified this PR in my own GPU, and it works properly.
Please merge this. Thanks for update!

PR #175, opened 2019-10-17
fix compilation

Roger-luo:
I'm not sure if you want this in master. This breaks the support of C++11 however


corryvrequan:
Thanks for PR!
As you guess, we update master branch only for new version releases. So, let me change target branch from master to dev. Since CI tests have passed, I would like to merge this branch to dev. After we check updated dev branch, we will merge dev branch to master.

PR #187, opened 2019-10-23
check avx2 support

corryvrequan:
The previous version's qulacs will crash when it is build under windows with CPU which does not support AVX2. I've added a cmake script to check whether AVX2 is supported or not, and if not disable SIMD optimization flags.


corryvrequan:
I've checked this PR works both for AVX2 compatible and not compatible machines. So, this is self-merged.

PR #200, opened 2019-11-30
Add `set -o pipefail` to fail CI when failing tests

y-yu:
- `.travis.ci` uses `if ....... fi` so CI may success even if tests fail
- So I add `set -o pipefail` when installing and testing
    - It comes from https://github.com/Instagram/IGListKit/pull/283


corryvrequan:
LGTM!

PR #202, opened 2019-12-16
Bugfix in trace preserving noise

corryvrequan:
In the previous version, probabilistic noise such as depolarizing noise etc are not trace preserving since it does not contains ingredients for Identity gate.
This is okay for state vector simulator, since Probabilistic gate treat CP-map with its probability less than unity as CPTP by compensating with Identity.
However, in the case of density matrix simulator, this is treated as trace decreasing map, and we shouldn't compensate with Identity map since we sometimes want to apply trace decreasing map.
Thus, I've updated definition of pre-implemented noise maps so that they become CPTP.


corryvrequan:
This is self-merged for next ver up

PR #204, opened 2019-12-26
Bugfix of bitstring sampling for density matrix

corryvrequan:
# summary

In the previous versions, sampling distribution of density matrix was wrong due to mistakenly used norm and abs. This bug was fixed, and simple test case was added.


corryvrequan:
Since this is an obvious bugfix, this is self-merged.

PR #207, opened 2019-12-30
Add CONTRIBUTING.md

y-yu:
- Qulacs build and test phases are turning complicated.
    - `make`, `mate test`, `make pythontest`, `python setup.py install` etc...
- So I add `CONTRIBUTING.md` which says the way how to bulid, format (#201 ) and test.
- I want to discuss if other things are needed to be written in `CONTRIBUTING.md` or not 🤔

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,44 @@
+# Contributing
+
+## Pull Request Process
+
+When you want to contribute to this project,
+please first discuss the change you wish to make via GitHub Issue.
+
+1. Check that your patch is legal under the [LICENCE](https://github.com/qulacs/qulacs/blob/master/LICENSE).
+2. Test in your computer.
+    - If you want to know how to do it please see `How to test` section below.
+3. PR to the `dev` instead of `master` branch.
+    - Because we use `master` branch as the release branch.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+$ ./script/build_msvc_2015.bat
+$ python setup.py install
+$ cmake --build ./visualstudio --target test --config Release
+$ cmake --build ./visualstudio --target pythontest --config Release
+```
+  
+### Tips
+
+- We only check the format in Linux
+  so Travis CI could point out some format errors even if there is nothing error in your Windows or macOS.
```
TenninYan: We only check the format in Linux.
Travis CI could point out format errors even if there are no errors in your Windows or macOS.

が正しい英語かなー？

y-yu: Would it be better to use `there is no error` instead of `there are no errors`?

TenninYan: そう変更してください！

y-yu: Sorry for my very very late fix 🙇

corryvrequan:
LGTM!

PR #208, opened 2020-01-12
Fix C++ gradient computing coefficients

corryvrequan:
In a previous PR, the definition of rotation Pauli was changed from exp(i \theta P) to exp(i\theta P/2).
However, gradient computing algorithm in vqcsim didn't updated according to this change.
I've doubled shifting rotation angles so that gradient computing becomes consistent.
This will affect who use template gradient computing algorithm with C++ qulacs.


corryvrequan:
This PR is considered to affect no other bugs, so this is self-merged.

PR #209, opened 2020-01-17
Add argument names and docstrings

corryvrequan:
I've added argument names to python binding.
Now you can see a help document of all the functions including constructors using "\_\_doc\_\_" member. For example,

```
import qulacs
print(qulacs.gate.CNOT.__doc__)
``` 
will output
```
CNOT(control: int, target: int) -> qulacs.QuantumGateBase

Create CNOT gate
```

When you do something in jupyter notebook, you can peek docstring by
1. Typing function (e.g.  "qulacs.gate.CNOT()"  )
2. Put cursor between parentheses
3. Type Shift+Tab


corryvrequan:
This is simple commend addition. This PR is self-merged.

PR #214, opened 2020-03-02
Add support for override of compiler optimization flags

kwkbtr:
Default compiler optimization flags can be overridden by OPT_FLAGS cmake variable.
It is also possible to override the flags via setup.py.
E.g.
```
python setup.py build_ext --opt-flags='-march=haswell -mtune=haswell -mfpmath=both'
```


kwkbtr:
The install script seems not working on Python 2.
https://travis-ci.org/qulacs/qulacs/jobs/657218534
I will look into it later.

corryvrequan:
I've checked the difference. I agree that adding build options as arguments is a nice solution.
Thanks for nice PR!

PR #219, opened 2020-04-23
Further SIMD optimization and thread number optimization for Pauli operations

corryvrequan:
# overview
This PR includes several independent updates for required optimizations.

## Refactor source codes
Since some codes in csim-folder are too long, I've separated them into small files for the readability.

## SIMD optimization for 2-qubit gate
In PR #181, several SIMD optimization codes are deployed to the public branch.
However, since 2-qubit gates require many conditions, they are left not SIMD optimized.
In this PR, 2-qubit gates are optimized with AVX2 instructions.

## Light quantum circuit optimizer
Previous quantum circuit optimizer performs detailed circuit optimization using commutation relations of quantum gates. Though this requires time scaling polynomially to the number of quantum gates, it takes a time comparable with quantum circuit simulation itself. Furthermore, typical Ansatz circuits only requires simple circuit optimization.
Thus, I've implemented light-weight-version quantum circuit optimizer, which lets gate A absorb gate B only if the following conditions are all satisfied.
1. gate A and B are neighboring
2. target qubits of gate B is subset of those of gate A
3. the number of target qubits of gate A is smaller than given max_block_size.

This optimization requires linear time to the number of gates.

## Tuning parallelization conditions
When quantum circuits are sufficiently small, overheads of parallelization exceed speed up by parallelization. In such a case, we should not do parallelization if qulacs is compiled with OpenMP flags.
Several basic gates are forced to single thread, but multi-qubit Pauli gates and its rotation are left un-optimized. I've tuned optimization condition for this gate.


corryvrequan:
Since 2-qubit gates are tested and other features affect nothing, this PR is self-merged.

PR #220, opened 2020-04-23
Add Toffoli and Fredkin

corryvrequan:
Toffoli and Fredkin gates are added to the collection of quantum gates.

Usage:
```python
from qulacs.gate import TOFFOLI, FREDKIN
#toffoli = TOFFOLI(control1, controls2, target)
#fredkin = FREDKIN(control, target1, target2)
```


corryvrequan:
Since this is a simple feature implementation, this is self-merged.

PR #221, opened 2020-04-23
Add diagonal gate

corryvrequan:
I've added a diagonal gate class for multiple qubits.

Usage
```
from qulacs.gate import DiagonalMatrix
target = [0,1]
diagonal_elem = [1,-1,-1,1]
gate = DiagonalMatrix(target, diagonal_elem)

from qulacs import StateVector
state = StateVector(3)
state.set_Haar_random_state(0)
```


corryvrequan:
Tested in my environment and seems no errors. This is self-merged.

PR #224, opened 2020-04-30
Update for clang compatibility

corryvrequan:
I've updated several small changes so that codes can be compiled with clang.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,5 @@
 
-cmake_minimum_required(VERSION 2.8)
-
-##### Check GCC version #####
-if(${CMAKE_C_COMPILER_VERSION} VERSION_LESS 7.0.0)
-	message(FATAL_ERROR, "gcc >= 7.0.0 is required.")
-endif()
-if(${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 7.0.0)
-	message(FATAL_ERROR, "g++ >= 7.0.0 is required.")
-endif()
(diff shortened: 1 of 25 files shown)


ValentinS4t1qbit:
Hello there ! I enjoy qulacs and find myself recommending it often, but the Mac installation is troublesome. Is there a reason why this branch has not been merged, or could we expect qulacs to install and work easily on Mac soon? (despite a potential decrease in performance due to clang ?)

I think being able to easily install qulacs on Mac, especially if you provide an easy way to do it with the actual GNU compilers, would be really useful !

vanyae-cqc:
@ValentinS4t1qbit I'm not sure if you still have this issue - You can install it on Mac (M1 chip) if you have gcc installed (eg via Homebrew). Just edit the `setup.py` script and change lines 87 and 88 to:

```
            cmake_args += ['-DCMAKE_C_COMPILER=/opt/homebrew/bin/gcc-11']
            cmake_args += ['-DCMAKE_CXX_COMPILER=/opt/homebrew/bin/g++-11'] 
```
(editing the path to wherever your gcc is installed)
And then run `python setup.py install`

Alternatively they could update to include gcc-11 in the `gcc_candidates` and `gxx_candidates`? Then you could set the `C_COMPILER` and `CXX_COMPILER` environment variables before running the install

forest1040:
qulacs-osakaとの統合でマージできなかった。。
clang対応を入れたいので、reopenしたい。
しかし、devブランチを削除したためreopenできない。

PR #225, opened 2020-05-16
fix: Dockerfile

FullteaOfEEIC:
When I used the original Dockerfile to create docker image, it failed tobuild. So I fixed it.


kwkbtr:
Thank you for contribution!

PR #229, opened 2020-06-10
Fix compiler specification in setup_gpu.py and setup_singlethread.py

kwkbtr:
The change applied to setup.py in #216 is also applied to setup_gpu.py and setup_singlethread.py.
Now diffs between the setup scripts are simple as follows:

```
$ diff setup.py setup_gpu.py
13c13
< project_name = 'Qulacs'
---
> project_name = 'Qulacs-GPU'
95a96
>         cmake_args += ['-DUSE_GPU:STR=Yes']
```

```
$ diff setup.py setup_singlethread.py
92a93
> 
95a97
>         cmake_args += ['-DUSE_OMP:STR=No']
```

### DIFF
--- a/setup_gpu.py
+++ b/setup_gpu.py
@@ -12,6 +12,14 @@ _VERSION = '0.1.9'
 
 project_name = 'Qulacs-GPU'
 
+def _is_valid_compiler(cmd):
+    try:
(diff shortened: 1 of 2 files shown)


corryvrequan:
LGTM!

PR #230, opened 2020-06-11
Fix compatibility with CUDA 10.1 and 10.2

ycchen1989:
This PR fix (with @leofang ) the following two issues:

1) The path to the NVCC: CUDA_NVCC_FLAGS (for CUDA 10.1 and 10.2)
2) Add the Conda GCC / G++ compiler to the list. This make it possible to install the package under the conda env with the compiler provided by conda.

### DIFF
--- a/setup_gpu.py
+++ b/setup_gpu.py
@@ -70,14 +70,14 @@ class CMakeBuild(build_ext):
             if env_gcc:
                 gcc_candidates = [env_gcc]
             else:
-                gcc_candidates = ['gcc', 'gcc-9', 'gcc-8', 'gcc-7']
+                gcc_candidates = ['gcc', 'gcc-9', 'gcc-8', 'gcc-7', 'x86_64-conda_cos6-linux-gnu-gcc']
             gcc = next(iter(filter(_is_valid_compiler, gcc_candidates)), None)
 
             env_gxx = os.getenv('CXX_COMPILER')
             if env_gxx:
                 gxx_candidates = [env_gxx]
             else:
-                gxx_candidates = ['g++', 'g++-9', 'g++-8', 'g++-7']
+                gxx_candidates = ['g++', 'g++-9', 'g++-8', 'g++-7', 'x86_64-conda_cos6-linux-gnu-g++']
             gxx = next(iter(filter(_is_valid_compiler, gxx_candidates)), None)
 
             if gcc is None or gxx is None:
@@ -86,6 +86,7 @@ class CMakeBuild(build_ext):
 
             cmake_args += ['-DCMAKE_C_COMPILER=' + gcc]
             cmake_args += ['-DCMAKE_CXX_COMPILER=' + gxx]
+            cmake_args += ['-DCUDA_NVCC_FLAGS=-I/usr/include']
 
             cmake_args += ['-DCMAKE_BUILD_TYPE=' + cfg]
             build_args += ['--', '-j2']


leofang:
Hi, as mentioned by @ycchen1989 we'd like to use Qulacs in a conda env. Please advise: 
1. whether the proposed change is acceptable, or you prefer a change at the CMakeLists level,
2. whether it is preferable that we simply pass conda's compiler name to `C_COMPILER`/`CXX_COMPILER`.

We plan to build and maintain a Qulacs package on Conda-Forge. To do so, we will first test the packaging process in [NSLS-II-Forge](https://github.com/nsls-ii-forge), a clone of the CF ecosystem for our internal facility use. The reason is that we have local CI set up to test GPU capabilities (whereas CF does not). Once we validate the whole process, we then submit a PR to https://github.com/conda-forge/staged-recipes to create a feedstock for Qulacs on CF.

Thanks.

corryvrequan:
Thanks for your PR, @ycchen1989 and @leofang !

We would like to support conda if possible, so this PR is welcome. Since I'm not familiar with conda and conda-forge ecosystems, and since this PR resolves two issues on CUDA and conda, let me ask a few things separately.

About a change of CUDA_NVCC_FLAGS

-  As far as we tested qulacs-GPU, there was no error without this change. Is this change required due to conda's feature, or required in all environments with CUDA10.1 and 10.2? We would like to know it for our future debug.

About change of compiler given by conda.

- As @leofang mentioned, we basically recommend passing a compiler name by setting <code>C_COMPILER</code> / <code>CXX_COMPILER</code> to avoid listing too many compiler names for specific purposes. Is this option not helpful for registering maintaining qulacs on conda?

leofang:
Hi @corryvrequan Thanks for your reply.
> About a change of CUDA_NVCC_FLAGS

Sorry we should have been clearer. The problem is not in conda, but in CUDA. In CUDA 10.1 & 10.2, the cuBLAS headers were moved from `/usr/local/cuda/include` to `/usr/include` under a horrible decision:
> With this release, on Linux systems, the cuBLAS libraries listed below are now installed in the /usr/lib/<arch>-linux-gnu/ or /usr/lib64/ directories as shared and static libraries. Their interfaces are available in the /usr/include directory:
> - cublas (BLAS)
> - cublasLt (new Matrix Multiply library)
>
> https://docs.nvidia.com/cuda/archive/10.1/cuda-toolkit-release-notes/index.html#cublas-new-features

And in 11.0 they are finally moved back, likely because it caused too much trouble:
> With this release, on Linux systems, the cuBLAS libraries listed below are now installed in the /usr/local/cuda-11.0 (./lib64/ for lib and ./include/ for headers) directories as shared and static libraries.
https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cublas-new-features

Now, if one always uses the system (default) compiler, it's likely that `/usr/include` is in the default include path, so everything would just work. However, if one uses non-default compiler (in our case, Anaconda's) then the problem reveals itself.

> About change of compiler given by conda

I think we might be able to set `C_COMPILER` / `CXX_COMPILER` in a conda recipe. I will do some experiments to confirm this. 

> We would like to support conda if possible

Thank you for your interest! How about this: I'll do the experiment on NSLS-II-Forge next week as I am more available by then. If everything goes smoothly, then I'll send a PR with the same recipe (likely with minor to no changes) to https://github.com/conda-forge/staged-recipes to create a qulacs-feedstock on Conda-Forge, and add you to the maintainer list of the recipe.

corryvrequan:
Thanks for your quick reply.

> About change of CUDA_NVCC_FLAGS

Oh, that's interesting... I understood the situation. Since this change seems helpful for several users and harmless to the other users, I agree with merging this change.

> About change of compiler given by conda
>> I will do some experiments to confirm this.

Thanks. I'm looking forward to comments or commits. 
After that, I would like to merge this PR.

> About sending recipes to conda-forge

We are happy if you could do so. Though I'm not so familiar with constructing conda recipes, I probably can review it if I'm added as a maintainer. Thanks for your huge support!

kwkbtr:
I happened to close this PR unintentionally.
I deleted `fix_gcc_version_detection_gpu` branch since it was already merged in #229, and then this PR was closed automatically because this PR uses the branch as its base.
Now I restored the branch and reopened this PR.
Could you change the base branch of this PR to `dev`?

leofang:
Please see https://github.com/nsls-ii-forge/qulacs-feedstock/pull/1#issuecomment-666081065 for a conda package. Let's revisit this PR later, as I currently don't have bandwidth to help polish it...

PR #239, opened 2020-06-26
Bugfix for parametric gate

corryvrequan:
In the previous release, `add_parametric_gate` function of `ParametricQuantumCircuit` class picks given gate without copying. This results in double-free and causes segmentation fault with the below codes for example.

```python
import qulacs
from qulacs.gate import ParametricRX
pqc = qulacs.ParametricQuantumCircuit(2)
prx = ParametricRX(0,0.1)
pqc.add_parametric_gate(prx)
del prx  # prx gate is released by python
del pqc  # prx gate in pqc is released by C++, and result in segfault.
```

This happens only when you add parametric rotation gates with `add_parametric_gate`, and does not happen with `add_parametric_RX_gate`, for example.
This PR fixes this issue, and I've added several tests on it. This issue is proposed by @akpc .

### DIFF
--- a/python/cppsim_wrapper.cpp
+++ b/python/cppsim_wrapper.cpp
@@ -391,17 +391,17 @@ PYBIND11_MODULE(qulacs, m) {
 		return ptr;
 	}, pybind11::return_value_policy::take_ownership, "Create state reflection gate", py::arg("state"));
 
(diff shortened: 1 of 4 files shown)


akpc:
LGTM!
ran test_qulacs.py
```
pqc = qulacs.ParametricQuantumCircuit(2)
prx = ParametricRX(0,0.1)
pqc.add_parametric_gate(prx)
```
bug fixed
Thanks for the quick fix!

PR #242, opened 2020-07-08
bugfix for measurement-log copy of GPU quantum state

corryvrequan:
QuantumStateBase::copy and its overrides must copy not only the state vector but also previous measurement results. It works for CPU state, However, in the case of GPU state, only state vector was copied and measurement results are not copied. I've fixed this point.


corryvrequan:
This is simple bugfix, so self merged.

PR #243, opened 2020-07-16
add CP map

corryvrequan:
I've added CP-map for the advanced use of an irreversible map.

```python
from qulacs.gate import CP, P0, H
p0 = P0(0)
# If true, state is normalized if Kraus operator is applied.
# If not, state is left unnormalized.
state_normalize = True 
# If true, probability distribution is normalized to unity.
# If not, no Kraus operator may be matched when the map is trace decreasing.
probability_normalize = False
# If true, return zero-filled vector when no Kraus operator is matched.
# If not, return the original input state when no Kraus operator is matched.
# This parameter is meaningless if you set probability_normalize = true.
assign_zero_if_not_matched = True

gate = CP([p0], state_normalize, probability_normalize, assign_zero_if_not_matched)

qs = QuantumState(1)
H(0).update_quantum_state(qs)
gate.update_quantum_state(qs)
```


corryvrequan:
This is checked by a person who proposed this issue. So, this is self-merged.

PR #244, opened 2020-07-31
Segfault in get parameter

corryvrequan:
In the previous version, get_parameter invokes segmentation fault.

In qulacs, if add_gate is called from python, copied gate instance is added to a quantum circuit to avoid double free from python and C++.
However, parametric_gate_list, which is an internal member variable to keep where a parametric variable is, is pointing a pointer before copy. Thus, it yields double free in several cases. I've fixed this problem.

This bug is reported by @kosukemtr .


corryvrequan:
This is reviewed by @kosukemtr, so this PR is merged.

PR #246, opened 2020-08-20
Fix C++ Example Compile Commands in Docs

kotamanegi:
This PR fix C++ example compile commands in document to work correctly.

I think C++ example compile commands written in document and readme is outdated, and does not work.
So I fixed example commands to work as expected.


corryvrequan:
Thanks, @kotamanegi . I also checked the updated command works.
Since almost all the users use python, many parts of C++ examples and tutorials may be outdated or unchecked. Sorry for inconvenience.

PR #247, opened 2020-08-22
Move OpenMP parallel directives of get_expectaiton_value

kosukemtr:
Moved openmp directive from `PauliOperator.get_expectation_value` to `GeneralQuantumOperator.get_expectation_value` for faster computation of expectation values.

This is done by defining new function `PauliOperator::get_expectation_value_single_thread` and single thread version of pauli expectation value functions in `csim/stat_ops_expectation_value`. `PauliOperator::get_expectation_value` is left as is to maintain the speed for single pauli expectation value.

### DIFF
--- a/src/cppsim/general_quantum_operator.cpp
+++ b/src/cppsim/general_quantum_operator.cpp
@@ -55,11 +55,19 @@ CPPCTYPE GeneralQuantumOperator::get_expectation_value(const QuantumStateBase* s
 		std::cerr << "Error: GeneralQuantumOperator::get_expectation_value(const QuantumStateBase*): invalid qubit count" << std::endl;
 		return 0.;
 	}
-    CPPCTYPE sum = 0;
-    for (auto pauli : this->_operator_list) {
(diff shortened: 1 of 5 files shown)


corryvrequan:
Thanks @kosukemtr, and sorry for very late reply!
When I try to merge this PR, I've found a problem in this patch. With this PR, single-thread version get_expectation_value is simultaneously called by OpenMP. However, when quantum states are allocated on GPU, parallelized call of updating function does not works. If you want to let this enhancement work correctly, enable OpenMP for CPU, and disable it for GPU. Can I ask you to check it works in both CPU and GPU?

kosukemtr:
Thanks very much for the review. I have modified the code. Now if the state is on GPU then the usual `get_expectation_value` is called without parallelizing.

corryvrequan (approved):

corryvrequan:
Sorry for very late review. I checked all the codes and they seems okay.
Thanks for speeding up this feature!

PR #248, opened 2020-09-03
Fix C++ library compile issues and guides

kotamanegi:
This PR fix the following two issues:
1. C++ library compile issues
In Eigen 3.3.5 with CUDA, a lot of Eigen users report that compile errors happen in Eigen/Half.h file.
This qulacs C++ library is being affected by this issue.
So, I tried updating version to Eigen 3.3.5 -> 3.3.7, which Eigen maintainers say that the issue has been removed.
I manually checked that this solution passes tests.

2. Usage guide of C++ with CUDA
When we use nvcc, we need to add many compile options compared to non-GPU qulacs.
Therefore, I think it is essential to add usage guide on documents.


corryvrequan:
Sorry for bothering you with out-of-date documents, and thanks for updating our documents.
The updated documents seem correct and nice!

PR #250, opened 2020-10-12
add probabilistic instrument

corryvrequan:
I've added a probabilistic instrument, which chose CP-map with given probability distribution, and logging what CP-map is applied.


corryvrequan:
This is an independent new feature. While it is already tested by me, it will be fixed if there is a bug.

PR #251, opened 2020-10-12
Fix memleak in observable

corryvrequan:
There was a memory leak in GeneralQuantumOperator::add_operator.
I've fixed this memory leak.


corryvrequan:
Since this fix of memory leak is small and seems okay, let me self-merge for now.

PR #252, opened 2020-10-26
Restore default thread number after disabling parallelization

corryvrequan:
Qulacs automatically disable parallelization when the number of qubits is small.
In the previous release, once parallelization is disabled in several functions, OMP_NUM_THREADS is overwritten to 1 and never restored to the original values. This does not affect the numerical results, but slow down the successive computation.
I've fixed this bug in this PR.

```
# Before PR
# num_qubit time_per_gate
25 0.07381165879113334
25 0.09755897521972656 Pauli
5 1.5777012765499026e-06
5 2.0947527406195033e-06 Pauli
25 0.19424397295171564
25 0.43433232307434083 Pauli

# After PR
# num_qubit time_per_gate
25 0.0742875646661829
25 0.10022400674365815 Pauli
5 1.5065100076042516e-06
5 1.933501226313842e-06 Pauli
25 0.08536406358083089
25 0.09836796351841517 Pauli
```


corryvrequan:
This passes tests, so self-merged.

Issue #257, opened 2020-11-06
Pauli string ends with whitespaces yields unexpected operator


r-imai-quantum:
# Symptom

```
$ python
>>> from qulacs import PauliOperator
>>> p = PauliOperator("X 0 Y 1", 1.0)
>>> assert len(p.get_index_list()) == 2
>>> p_wrong = PauliOperator("X 0 Y 1 ", 1.0) # Pauli string ends with whitespace
>>> assert len(p_wrong.get_index_list()) == 3 # user expects 2 here
```

r-imai-quantum:
This is fixed by #276

PR #258, opened 2020-11-06
Fix: ignore trailing whitespaces in Pauli string.

r-imai-quantum:
This PR fixes issue #257.

### Added

- `rtrim` function to `utility`.

### Changes

- Trim trailing whitespaces in given Pauli string in the beginning of process to construct PauliOperator.
 Then, PauliOperator will be created properly.

### How it works

```
Python 3.7.8 (default, Aug 17 2020, 14:49:28)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from qulacs import PauliOperator
>>> p = PauliOperator("X 0 Y 1    ", 1.0) # with writespaces
>>> assert len(p.get_index_list()) == 2
```

### DIFF
--- a/src/cppsim/pauli_operator.cpp
+++ b/src/cppsim/pauli_operator.cpp
@@ -31,7 +31,7 @@ extern "C"{
 
 PauliOperator::PauliOperator(std::string strings, CPPCTYPE coef){
     _coef = coef;
-    std::stringstream ss(strings);
+    std::stringstream ss(rtrim(strings));
     std::string pauli_str;
     UINT index, pauli_type=0;
     while(!ss.eof()){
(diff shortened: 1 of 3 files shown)


corryvrequan:
Thanks for PR!

CI says there is an error for compile with MSVC. According to errors, we need `#include <cctype>` for using std::isspace, which seems implicitly declared in gcc without including cctype. So, please add `#include <cctype>` for utility.cpp.

r-imai-quantum:
Thank you for your suggestion!
I was a little confused by the difference between gcc and MCVC.
I've updated codes, and all tests are passed now. 😄

corryvrequan:
Thanks for fixing it, and very sorry for late review.
Thanks for contribution!

PR #261, opened 2020-11-27
Fix CI on mac; upload built wheel files as CI artifacts

kwkbtr:
In GitHub Actions virtual environment [macos-10.15/20201115.1](https://github.com/actions/virtual-environments/blob/macOS-10.15/20201115.1/images/macos/macos-10.15-Readme.md), linking some functions from libgomp fails since the linker cannot find some library files from gcc-8 installed via Homebrew.
Failure example: https://github.com/qulacs/qulacs/runs/1462123962?check_suite_focus=true#step:10:165
Added a command that fixes this issue by (brew) re-linking files from gcc-8.
cf. https://stackoverflow.com/a/55500164

Also added a setting to upload built wheel files as CI artifacts so that we can download and test them on local machines before publishing to PyPI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,11 @@ jobs:
       - name: Run cibuildwheel
         run: python -m cibuildwheel --output-dir wheels
 
+      - name: Upload wheel files as build artifacts
+        uses: actions/upload-artifact@v2
+        with:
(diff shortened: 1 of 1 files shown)


kwkbtr:
@y-yu Please review 🙏

y-yu (approved):
LGTM

PR #262, opened 2020-11-27
CI: Publish qulacs-gpu sdist to PyPI

kwkbtr:
-  Sync setup.py for GPU and single thread with the default setup.py
  - They should have been updated in #233
- Add CI setting to upload sdist for GPU enabled version
- Install twine only when publishing to PyPI


y-yu (approved):
I don't understand who uses `QULACS_OPT_FLAG` environment variable but it seems to be OK.

y-yu (approved):

corryvrequan (approved):

PR #266, opened 2020-11-28
Use latest cibuildwheel

kwkbtr:
The old version (1.4.2) of cibuildwheel does not build wheel files for Python 3.9.
Let's use the latest cibuildwheel so that wheel files for new Python versions are automatically built.


y-yu (approved):
We uses fixed version of `cibuildwheel` under current CI setting. It's more stable to reproduce CI results but there's no serious problem to use the latest version, I suppose.

PR #267, opened 2020-11-29
Set API tokens for PyPI in CI setting

kwkbtr:
API tokens used for upload to PyPI are set from GitHub secrets.
Qulacs and Qulacs-GPU use different tokens, so each token needs to be set individually in each upload job.


kwkbtr:
I'm merging this without waiting for CI to finish since it should not change behavior of CI jobs other than uploading jobs, which are executed only when the CI is triggered by a tag.

PR #274, opened 2020-12-07
Update readme with correct build command

corryvrequan:
I've updated out-dated or wrong directions of qulacs build commands in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ A binary that is installed via pip command is optimized for Haswell architecture
 - C++ compiler (gcc or VisualStudio)
     - gcc/g++ >= 7.0.0 (checked in Linux, MacOS, cygwin, MinGW, and WSL)
     - Microsoft VisualStudio C++ 2015 or later
-- python 2.7 or 3.x
+- Python 2.7 or 3.x
 - cmake >= 3.0
 - git
 - (option) CUDA >= 8.0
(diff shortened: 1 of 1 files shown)


y-yu:
This PR contains the merge commit (https://github.com/qulacs/qulacs/pull/274/commits/987474b31a6e60eba116d2e40ab538dcf8086038) so it would be better to `git rebase origin/master` and `git push origin HEAD -f` maybe.

README.md
```diff
@@ -212,12 +120,101 @@ int main(){
 }
 ```
 
-Build command for g++:
+
+## Install Python library from source
+
+To install Qulacs optimized for your system, we recommend the following install procedure for faster simulation of quantum circuits, while this requires a compiler and takes time for installation.
+
+A binary that is installed via pip command is optimized for Haswell architecture. Thus, Qulacs installed via pip command does not work with a CPU older than Haswell. If your CPU is newer than haswell, Qualcs built from source shows the better performance.
```
y-yu, line 128:
```suggestion
A binary that is installed via pip command is optimized for Haswell architecture. Thus, Qulacs installed via pip command does not work with a CPU older than Haswell. If your CPU is newer than Haswell, Qualcs built from source shows the better performance.
```

README.md
```diff
@@ -212,12 +120,101 @@ int main(){
...
+
+### Requirement
+
+- C++ compiler (gcc or VisualStudio)
+    - gcc/g++ >= 7.0.0 (checked in Linux, MacOS, cygwin, MinGW, and WSL)
+    - Microsoft VisualStudio C++ 2015 or later
+- python 2.7 or 3.x
```
y-yu, line 135:
```suggestion
- Python 2.7 or 3.x
```

Issue #259, opened 2020-11-06
Pauli string ends with whitespaces yields unexpected operator


tsuvihatu:

tsuvihatu:
PauliOperator doesn't have get_qubit_count method?

forest1040:
This code was misstake this point.
PauliOperator was more QuantumState size.
These day, we fixed validation.

```
>>> from qulacs import PauliOperator, QuantumState 
>>> n = 2 
>>> coef = 2.0 
>>> Pauli_string = "X 0 X 1 X 3" 
>>> pauli = PauliOperator(Pauli_string,coef) 
>>>      
>>> state = QuantumState(n) 
>>> state.set_Haar_random_state() 
>>> value = pauli.get_expectation_value(state) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Error: PauliOperator::get_expectation_value(QuantumStateBase*):The number of qubit in PauliOperator is greater than QuantumState.PauliOperator: 4 QuantumState: 2

```

PR #276, opened 2020-12-21
Fix pauli operator get expection value

kotamanegi:
This PR fixes issue #257 and #259.

# issue #257
This issue says that wrong Pauli Strings is processed without errors.
This happens because initialization process in parsing Pauli Strings is broken.
To prevent this, I've added appropriate initialization process.

# issue #259 
This issue says that PauliOperator::get_expection_value() will broke when number of qubits in PauliOperator is larger than in QuantumState.
To prevent this, I've added PauliOperator::get_qubit_count() function and do some checks before applying PauliOperator using added function.
 
# Other minor changes

Added Tests for PauliOperator.
Refactored PauliOperator::get_expection_value() GPU code.


corryvrequan:
LGTM. Sorry for late merge.

PR #277, opened 2021-01-22
Disable some options when the target arch. is not x86

r-imai-quantum:
Some default options passed to the compiler are unavailable in architectures other than x86.
This pull request disables these options when the host system is not x86.

I confirmed that, after this change, the compile process becomes to work well in my chromebook (aarch64 architecture).

# Changes 

- If the architecture only matches `^x86.*`, SIMD and `-mfpmath=both` (this enables SSE) are enabled.
- Some macro IFs are added to prevent unused header files to be loaded when SIMD is disabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,12 +11,17 @@ endif()
 
 
 ##### Set default behavior #####
-set(DEFAULT_USE_SIMD Yes)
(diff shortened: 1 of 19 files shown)


corryvrequan:
Sorry for late comments.
LGTM. Thanks for contribution!

PR #283, opened 2021-04-19
Add feature of several density matrix controls

corryvrequan:
I've added `tensor_product`, `permutate_qubit` and `partial_trace` for density matrix controls, and added tests for them.
I also fixed tests for openfermion text to allow an epsilon error.


corryvrequan:
Since this is an implementation of new features, let me self-merge.

PR #285, opened 2021-07-13
Add light partial trace

corryvrequan:
When we convert a state vector to a density matrix, the state vector is temporally converted to a density matrix with the same number of qubits. I've updated `partial_trace` function not to use a temporal large density matrix.
Now we can treat partial trace from a 20-qubit state vector to 10-qubit density matrix with a reasonable memory size.


corryvrequan:
Github action script for MacOS seems outdated. I've updated script to verify gcc-8 is installed.

corryvrequan:
This PR is self-merged. Fixed if issue will be reported.

PR #286, opened 2021-08-06
Update SIMD options

corryvrequan:
I've changed if-condition to check the host processor supports the SIMD instruction and option "-mfpmath=both".
Previous condition only supports processor that returns "x86_64", but some processor returns "AMD64".
I've updated condition and it works in recent processors.


r-imai-quantum (approved):
LGTM!

Issue #292, opened 2021-11-10
Noisy circuit memory leak python Mac OS


JamesB-1qbit:
There appears to be a memory leak when building a circuit with noise from python in Mac OS.

Example code to reproduce
```python
import qulacs
from qulacs.gate import DepolarizingNoise

def create_circuit():
    target_circuit = qulacs.QuantumCircuit(6)
    qulacs.QuantumCircuit.add_H_gate(target_circuit, 0)
    qulacs.QuantumCircuit.add_H_gate(target_circuit, 1)
    qulacs.QuantumCircuit.add_H_gate(target_circuit, 2)
    qulacs.QuantumCircuit.add_H_gate(target_circuit, 3)
    qulacs.QuantumCircuit.add_H_gate(target_circuit, 4)
    qulacs.QuantumCircuit.add_H_gate(target_circuit, 5)
    qulacs.QuantumCircuit.add_CNOT_gate(target_circuit, 0, 1)
    qulacs.QuantumCircuit.add_CNOT_gate(target_circuit, 1, 2)
    qulacs.QuantumCircuit.add_CNOT_gate(target_circuit, 2, 3)
    qulacs.QuantumCircuit.add_CNOT_gate(target_circuit, 3, 4)
    qulacs.QuantumCircuit.add_CNOT_gate(target_circuit, 4, 5)
    target_circuit.add_gate(DepolarizingNoise(0, 0.5))

    return target_circuit

for i in range(200000):
    circuit = create_circuit()
```

running the program with `memory_profiler` i.e. `mprof run python3 memleak_qulacs.py` results in `mprof plot`
![qulacs_memleak](https://user-images.githubusercontent.com/84878946/141201226-2fbc2280-3eb7-4c55-8851-93daf8f92130.png)

If you remove the `DepolarizingNoise` gate. There is no memory leak. The leak is also present for `TwoQubitDepolarizingNoise` and `Probabilistic` type gates.

corryvrequan:
Hi @JamesB-1qbit .
I have checked there are memory leak bugs, as you pointed out.
Now I have fixed the bugs and test passes. The bugfix will be merged to the dev branch and fixed from the next release.

The following figure is the results of your codes with the fixed version.
![bench](https://user-images.githubusercontent.com/44130003/141496544-aaad2536-390a-49f0-b119-d6fb19f59148.png)

Thanks for reporting this issue!

JamesB-1qbit:
Thanks so much!

PR #293, opened 2021-11-12
Fix memory leak in composite gates

corryvrequan:
In the previous release, there was a memory leak in composite gate such as `Probabilistic` and `CPTP`.
This was reported in #292. The memory leaks are fixed.


corryvrequan:
I've removed python 2.7 from CI. I updated visual studio from 2017 to 2019.
Since windows + python3.10 cannot be built, I removed windows python 3.10 from cibuildwheel.

PR #307, opened 2022-04-01
Change file encoding from ShiftJIS to UTF-8

kwkbtr:


y-yu (approved):
It would be more better to translate these Japanese comments to English later.

corryvrequan (approved):

Issue #314, opened 2022-06-21
Expectation value can be wrong when using DensityMatrix (Severe bug)


doomspec:
Here is a minimal reproducing code I made.
```
from qulacs import DensityMatrix, QuantumState
from qulacs import QuantumCircuit, Observable

n_qubit = 1

operator = Observable(n_qubit)
operator.add_operator(1, "Y 0")
operator.add_operator(1, "Z 0")

circuit = QuantumCircuit(n_qubit)
circuit.add_RX_gate(0, 2.0)

state = DensityMatrix(n_qubit)
circuit.update_quantum_state(state)
dm_exp = operator.get_expectation_value(state)
print(dm_exp)

state = QuantumState(n_qubit)
circuit.update_quantum_state(state)
qs_exp = operator.get_expectation_value(state)
print(qs_exp)

```
Running the code will give
```
-1.325444263372824
0.4931505902785395
```
which does not match. 

I guess this bug is caused by the simulator giving different sign on Pauli Y. The version of python I am using is Python 3.9.

corryvrequan:
Hi, @doomspec 
I reproduced this bug with your code, and I have fixed this problem in PR #315.
This happens when an observable contains a term with an odd number of Pauli-Y and evaluates density matrices, i.e., the sign of Pauli-Y operators is effectively flipped in the calculation of density matrices. While we expected that these features are intensively tested by calculating random Pauli operations and with an example of open-fermion, these cases are unexpectedly leaked from the test cases.

Since this is a basic and critical one, we plan to update the main branch soon and announce it in the release. 
We are very sorry for being not aware of such a basic one, and thanks a lot for reporting a critical bug in Qulacs.

kwkbtr:
The new vesion v0.3.1 with the bug fix is just released.
https://github.com/qulacs/qulacs/releases/tag/v0.3.1

PR #315, opened 2022-06-21
Fix a bug in the calculation of expectation values on odd-number Pauli-Y observable for density matrix

corryvrequan:
In the previous version, the sign of observable evaluation for density matrices is flipped when it contains a term with an odd number of Pauli-Y operators, such as "Y 0" or "Z 0 Y 1". Although the expectation-value calculations are tested with random Pauli operations and a specific open-fermion text, this bug happens since random tests are not random due to another bug, and an open-fermion example contains an even number of Pauli-Y operators. 
I have replaced the calculation function of expectation values for density matrices with a stable one and fixed the bug in the tests of random observable. This bug is reported in #314 and is expected to be included from the first implementation of density matrices. We are very sorry for being unaware of the bug in such a fundamental part.


corryvrequan:
Since this bug patch is critical, fixed tests passed, and main branch will be updated soon, this PR is self-merged.

PR #334, opened 2022-09-21
Refactoring NoiseSimulator/GradCalculator

kotamanegi:
NoiseSimulator/GradCalculator is poorly written, so in this PR I've refactored variable name & deleted unused codes.
Scrap code is in the benchmark/ folder, so I removed them.
This change should have no changes to user interfaces.


forest1040:
I think the code is now easy to understand.
Especially the delta constant!
I'm not sure if the benchmark code is needed or not, but I think I can still leave it there without deleting it.
Rewrite the benchmark code and then delete it?

KowerKoint (approved):
LGTM.
Some parts have been more easy to read.

I do not know whether codes in `benchmark/`is needed, so you may delete it if it is absolutely useless.

kotamanegi:
Deleted files in `benchmark/` is totally useless and is even uncompilable now.
It's rather easy to write benchmark code from scratch than making benchmark code works.
Since we're using git, we can restore files using `git log`.
So, it's way better to delete unused `benchmark/` files.

forest1040:
OK! I see. Understood. Please delete it.

forest1040 (approved):

Issue #344, opened 2022-10-06
QuantumCircuitSimulator terminates abnormally.


forest1040:
Even a code that does nothing as follows will terminate abnormally.

```
from qulacs import QuantumState, QuantumCircuit, QuantumCircuitSimulator, Observable
n = 3
state = QuantumState(n)

circuit = QuantumCircuit(n)
for i in range(n):
    circuit.add_H_gate(i)

sim = QuantumCircuitSimulator(circuit, state)
```

```
$ python sim.py
zsh: segmentation fault  python sim.py
```

ikanago:
I reproduced the same error with C++ version, even without adding H gate.
```cpp
#include "cppsim/circuit.hpp"
#include "cppsim/observable.hpp"
#include "cppsim/simulator.hpp"
#include "cppsim/state.hpp"

int main() {
    const unsigned int n = 3;
    auto state = QuantumState(n);
    QuantumCircuit circuit(n);
    const auto simulator = QuantumCircuitSimulator(&circuit, &state);
}
```
Error emitted:
```
munmap_chunk(): invalid pointer
[1]    14717 abort      ./a.out
```
So this should be caused by the constructor of `QuantumCircuitSimulator`, not Python binding layer or others.

ikanago:
Deleting following 3 `delete`s, the error disappeared.
https://github.com/qulacs/qulacs/blob/ddd6c8f38cbaf8c7a4bb72ced495d16e75117145/src/cppsim/simulator.cpp#L15-L19

This is because `delete`ing the pointer is not the responsibility of the class, but a scope which defined the pointee.

forest1040:
https://github.com/qulacs/qulacs/pull/364

PR #364, opened 2022-10-18
#344: QuantumCircuitSimulator terminates abnormally.

forest1040:
Fix to delete _buffer and _state if needed.


src/cppsim/simulator.cpp
```diff
@@ -9,12 +9,17 @@
...
+    }
+    // _circuitはQuantumCircuitSimulatorを継承しているParametricQuantumCircuitSimulatorが
+    // circuitのポインタを共有しておりtestで最適化した回路を使用するので、
+    // 渡されたcircuitをそのまま使用する必要がある。
 };
 
 QuantumCircuitSimulator::~QuantumCircuitSimulator() {
-    if (_circuit != NULL) delete _circuit;
-    if (_state != NULL) delete _state;
+    if (_own_state) delete _state;
```
ikanago, line 22: `_state` points to the `QuantumStateBase` which is not owned by this class, so this destructor should not  `delete` the `_state`. This should be deleted at the scope where the state is initialized.

forest1040: I had the wrong code.
https://github.com/qulacs/qulacs/blob/fix-344-simulator-abort/src/cppsim/simulator.cpp#L12
The check here was intended to ensure that the initial_state was NULL.
I'll fix it.

forest1040: Fixed checking of initialization parameters.
if initial _ state is null, new _state.
And delete destructor.

forest1040:
Modified the destructor to check what it frees, since it swap_state_and_buffer() when it is executed.
and add simurator test code.

src/cppsim/simulator.hpp
```diff
@@ -17,6 +17,8 @@ class DllExport QuantumCircuitSimulator {
     QuantumCircuit* _circuit;
     QuantumStateBase* _state;
     QuantumStateBase* _buffer;
+    bool _own_state = false;
+    bool _is_swap = false;
```
ikanago, line 21: We should have more direct name like `_is_buffer_used`, shouldn't we?

src/cppsim/simulator.cpp
```diff
@@ -61,5 +70,9 @@ void QuantumCircuitSimulator::swap_state_and_buffer() {
         _buffer = new QuantumState(_state->qubit_count);
         _buffer->set_zero_state();
     }
-    std::swap(_state, _buffer);
+    _is_swap = !_is_swap;
+    auto tmp = new QuantumState(_state->qubit_count);
```
ikanago, line 74: `tmp` should be released in this scope.

ikanago (changes requested):

src/cppsim/simulator.hpp
forest1040: It was. I corrected it because the swap () was wrong, so if the _buffer was reserved, it needed to be opened.
I'll fix it.

src/cppsim/simulator.cpp
forest1040: Thx!

forest1040:
Fixed it.

src/cppsim/simulator.hpp
ikanago: Now we don't need `_is_swap`.

ikanago (changes requested):

forest1040:
> Now we don't need _is_swap.

Oh! Thank you. I fixed.

ikanago (approved):
LGTM 👍

PR #401, opened 2022-11-10
#396 Add CI and FindSVE.cmake script to support SVE

m-ymzk:
Support 2 items of #396 
- Add CI using QEMU to test this code for aarch64

  Unfortunately, I couldn't find a way to use these commands in cross-compilation.
  - ccache
  - CHECK_C_SOURCE_RUNS
  - CHECK_CXX_SOURCE_RUNS
  - python / venv

- Add cmake_script/FineSVE.cmake : As with the code of AVX2 SIMD, do not add build options, but enable them if the environment has a sve instruction at build time.


kotamanegi (approved):
I've checked CI result and it seems working. LGTM :+1:

forest1040 (approved):
LGTM!

PR #402, opened 2022-11-11
Disable gcov by default

ikanago:
Compiling and running test with gcov takes longer time than usual.
So disabled gcov by default as we do not need check coverage except on CI.


forest1040 (approved):
Thank you!

KowerKoint:
By this change, `COVERAGE=No` is a default option now, so we cannot get correct coverage on current VSCode task.
I run `test and generate coverage` task in VSCode and watch `coverage/html` on branch, I only got coverage of global function.
A new task to run `USE_TEST=Yes COVERAGE=Yes script/build_gcc.sh` should be placed on `.vscode/tasks.json` and used for `test and generate coverage` task.
For the same reason, Ubuntu CI should run `build_gcc.sh` with `COVERAGE=Yes`.

ikanago:
I fixed them as you suggested!

KowerKoint:
Thank you.
Then, the dependency of `test and generate coverage` task should be `build gcc with coverage option` .

KowerKoint (approved):

kotamanegi (approved):
LGTM:+1:

PR #447, opened 2022-12-19
Revert "The universal2 setting was restored."

forest1040:
Revert because universal2 will not be enabled.


KowerKoint (approved):
OK
universal2 wheel seems to be just a complicated thing because there are both x86_64 and one and arm64 one.

PR #448, opened 2022-12-19
Revert "add darwin CMAKE_OSX_ARCHITECTURES"

forest1040:
Reverts qulacs/qulacs#439

In the current CI environment, because MAC is using gcc -10,
Revert because it does not support ARM 64 and results in an error.


KowerKoint (approved):
:+1:

PR #451, opened 2022-12-20
specify type for lambda arguments

### LINKED ISSUE #450: specify type in lambda arguments
This fix will resolve https://github.com/qulacs/qulacs/pull/446#issuecomment-1354047050

KowerKoint:
close #450
I think using "const reference" if possible is better for performance, so I used that when resolving `auto`.


forest1040 (approved):
👍

kotamanegi (approved):
:+1:

ikanago (approved):
👍

PR #452, opened 2022-12-20
397: Supports SPMD-type parallelization with MPI

m-ymzk:
#397 : Supports SPMD-type parallelization with MPI

## content of proposal
- [x] Add CI to build with mpicc/mpicpp (for linux)
- [x] Add USE_MPI build option
- [x] Add a new class module to comunication with MPI
- [x] Add a new QuatnumState API to divide the State Vector and to distribute it across multiple processes
- [x] Support QuantumState.set_computational_basis(), set_Haar_random_state()
- [x] Support QuantumState:sampling()
- [x] Support QuantumState:copy(), load()
- ~~Change behavior when specifying random number seed to deterministic (Singletonization of the Random class)~~
- [x] Corresponds to the thread number specified by the environment variable
- [x] Support named gates
  - [x] I, X, Y, Z
  - [x] H, S, T, Sdag, Tdag, sqrtX, sqrtXdag, sqrtY, sqrtYdag, P0, P1
  - [x] RX, RY, RZ
  - [x] CNOT, SWAP
  - [x] CZ
- [x] Support general-purpose gates
  - [x] single-target diagonal matrix gate
  - ~~multi-target diagonal matrix gate~~
  - ~~multi-control multi-target diagonal matrix gate~~
  - [x] single-target dense matrix gate
  - [x] double-target dense matrix gate
  - [x] multi-target dense matrix gate (#target <= #local_qubit)
  - [x] single-control single-target dense matrix gate
  - [x] multi-control single-target dense matrix gate
  - [x] single-control multi-target dense matrix gate
  - [x] multi-control multi-target dense matrix gate
- [x] Support get_expectation()


python/cppsim_wrapper.cpp
```diff
@@ -1129,6 +1137,17 @@ PYBIND11_MODULE(qulacs_core, m) {
...
+#if 0  // not supported yet
+        .def("update_quantum_state",
+            (void (QuantumCircuit::*)(QuantumStateBase*, UINT)) &
+                QuantumCircuit::update_quantum_state,
+            "Update quantum state with random seed", py::arg("state"),
+            py::arg("seed"))
+        .def("update_quantum_state",
+            (void (QuantumCircuit::*)(QuantumStateBase*, UINT, UINT, UINT)) &
+                QuantumCircuit::update_quantum_state,
+            py::arg("state"), py::arg("start"), py::arg("end"), py::arg("seed"))
```
KowerKoint, line 1149: We do not add multiple comments for overloaded function ( https://github.com/Qulacs-Osaka/qulacs-osaka/issues/292 ).
We use `py::overload_cast ( https://github.com/qulacs/qulacs/pull/386 ).

src/cppsim/state.hpp
```diff
@@ -36,21 +41,76 @@ class QuantumStateBase {
...
+        }
+        UINT log_nodes = std::log2(mpisize);
+        if (use_multi_cpu &&
+            (qubit_count_ >= (log_nodes + 2))) {  // minimum inner_qc=2
+            this->_inner_qc = qubit_count_ - log_nodes;
+            this->_outer_qc = log_nodes;
+        } else {
+            this->_inner_qc = qubit_count_;
+            this->_outer_qc = 0;
+        }
```
KowerKoint, line 95: if qubit_count=6 and log_nodes=6, it seems to be inner_qc=6 and outer_qc=0. Could you tell me why you do not let it to be inner_qc=2 and outer_qc=4?

m-ymzk: First, one reason is the limitations of current implementations. A gate update operation with multiple targets are only supported if all targets are local. As the qubit number of commonly used gates, 2 is specified. If a generic gate with more than the number of local qubits is used, a Not Implemnted exception is raised. https://github.com/qulacs/qulacs/blob/fe2a8d68a71171f28e83cafa8469cd4ab1e54db5/src/csim/update_ops_matrix_dense_multi.cpp#L142-L143

Secondly, for practical reasons, there is little advantage in using distributed placement when the local qubit is smaller than for example 10. The advantage of distributed placement is that it deals with a state vector of qubits that is larger than it holds locally. Having around 20-30 qubits locally also allows the parallel performance of the arithmetic unit and/or memory bandwidth to be exploited.

KowerKoint: I see. Thank you.
Then, isn't it better place all the qubits in local when qubit_count is small enough?
When qubit_count=6, why do you distribute qubits only if log_nodes=1,2,3,4?

m-ymzk: Good question!
There is certainly no need for a distributed placement as there is no benefit in a small qubit.
The only reason for this is to test the operation in a distributed placement with a small qubit.

Probably, it is possible to set this threshold in an environment variable(ex. 10 ~ 20).
However, as long as the user is instructed to use_multi_cpu=True, it would be natural for the operation to follow that direction as much as possible.

src/cppsim/state.cpp
```diff
@@ -19,6 +19,11 @@ CPPCTYPE inner_product(
 }
```
KowerKoint, line 19: I think `state::inner_product()` should call `state_inner_product_mpi()` when `outer_ac > 0`

KowerKoint:
Mark as NotImplemented for some function in `state_dm.hpp/cpp`  using `QuantumStateCpu` if `outer_qc > 0`.

- `DensityMatrix::set__Haar_random_state(UINT)` (force not to use MPI)
- `DensityMatrix::load(const QuantumStateBase*)
- `state::partial_trace(const QuantumStateCpu*, std::vector<UINT>)`
- `state::make_mixture(CPPCTYPE, const QuantumStateBase*, CPPCTYPE, const QuantumStateBase*)

python/cppsim_wrapper.cpp
m-ymzk: I removed it with fe2a8d6

src/cppsim/state.cpp
m-ymzk: Right! 
I provided partial support to the inner product in fe2b582.
(Calculating the inner product of the variance and non-dispersion vectors is a TODO.)

m-ymzk: inner product with multi-cpu is now supported.

src/csim/update_ops_control_single_target_single.cpp
```diff
@@ -251,3 +252,136 @@ void single_qubit_control_single_qubit_dense_matrix_gate_simd(
...
+    UINT index_offset);
+void single_qubit_control_single_qubit_dense_matrix_gate_mpi_OO(
+    CTYPE* t, const CTYPE matrix[4], CTYPE* state, ITYPE dim, int flag);
+
+void single_qubit_control_single_qubit_dense_matrix_gate_mpi(
+    UINT control_qubit_index, UINT control_value, UINT target_qubit_index,
+    const CTYPE matrix[4], CTYPE* state, ITYPE dim, UINT inner_qc) {
+    MPIutil& m = MPIutil::get_inst();
+    const UINT rank = m.get_rank();
+    const UINT control_rank_bit = 1 << (control_qubit_index - inner_qc);
```
KowerKoint, line 269: Behavior of shifting by negative value is undefined.
Calculating those in if blocks is safer.

src/csim/update_ops_matrix_dense_double.cpp
```diff
@@ -1700,5 +1701,65 @@ void double_qubit_dense_matrix_gate_sve(UINT target_qubit_index1,
...
+            SWAP_gate_mpi(target_qubit_index2, act_target_qubit_index2, state,
+                dim, inner_qc);
+
+            double_qubit_dense_matrix_gate_c(target_qubit_index1,
+                act_target_qubit_index2, mat, state, local_dim);
+
+            SWAP_gate_mpi(act_target_qubit_index2, target_qubit_index2, state,
+                dim, inner_qc);
+        }
+    }
```
KowerKoint, line 1763: In current case, there is at least one outer_target when `double_qubit_dense_matrix_gate_mpi()` is called, but if it is called with local targets in the future, `SWAP_gate_mpi()` with same targets are called and get trouble.

src/csim/update_ops_matrix_dense_multi.cpp
```diff
@@ -121,3 +126,70 @@ void multi_qubit_dense_matrix_gate_parallel(const UINT* target_qubit_index_list,
...
 }
+
+#ifdef _USE_MPI
+void multi_qubit_dense_matrix_gate_mpi(const UINT* target_qubit_index_list,
+    UINT target_qubit_index_count, const CTYPE* matrix, CTYPE* state, ITYPE dim,
+    UINT inner_qc) {
+    UINT num_outer_target = 0;
+    for (UINT i = 0; i < target_qubit_index_count; ++i)
+        if (target_qubit_index_list[i] >= inner_qc) ++num_outer_target;
+
```
KowerKoint, line 137: Throw exception if num_outer_target > inner_qc.

src/csim/update_ops_control_single_target_multi.cpp
```diff
@@ -63,3 +68,81 @@ void single_qubit_control_multi_qubit_dense_matrix_gate(
...
+    UINT* new_target_index_list =
+        (UINT*)malloc((size_t)(sizeof(UINT) * target_qubit_index_count));
+
+    // count outer-target qubit
+    UINT num_outer_target = 0;
+    for (UINT i = 0; i < target_qubit_index_count; ++i) {
+        new_target_index_list[i] = target_qubit_index_list[i];
+        if (target_qubit_index_list[i] >= inner_qc) ++num_outer_target;
+    }
+
```
KowerKoint, line 86: Throw exception if num_outer_target > inner_qc.

src/csim/update_ops_control_multi_target_multi.cpp
```diff
@@ -67,3 +72,127 @@ void multi_qubit_control_multi_qubit_dense_matrix_gate(
...
+        new_control_index_list[j] = control_qubit_index_list[j];
+        new_control_value_list[j] = control_value_list[j];
+    }
+    // count outer-target qubits
+    UINT num_outer_target = 0;
+    for (UINT i = 0; i < target_count; ++i) {
+        new_target_index_list[i] = target_qubit_index_list[i];
+        if (target_qubit_index_list[i] >= inner_qc) ++num_outer_target;
+    }
+
```
KowerKoint, line 99: Throw exception if num_outer_target > inner_qc.

src/csim/update_ops_control_single_target_multi.cpp
m-ymzk: Thank you for pointing it out. I add an exception like the one above.

src/csim/update_ops_control_multi_target_multi.cpp
m-ymzk: Thank you. I added an exception here as well.

src/csim/update_ops_control_single_target_single.cpp
m-ymzk: Fixed and a little refactored at 8c0ad7c

m-ymzk:
sorry for my late reply.

- DensityMatrix::set__Haar_random_state(UINT) (force not to use MPI)

Currently, it is not made with multi-cpu. default is not to use multi-cpu https://github.com/qulacs/qulacs/blob/8a3d98088a936145752cc952546ddba6718ffa2e/src/cppsim/state_dm.hpp#L74

- DensityMatrix::load(const QuantumStateBase*)
- state::partial_trace(const QuantumStateCpu*, std::vector<UINT>)
- state::make_mixture(CPPCTYPE, const QuantumStateBase*, CPPCTYPE, const QuantumStateBase*)

I added NotImplementedException in state_dm functions at 17d544e

src/csim/update_ops_control_single_target_single.cpp
```diff
@@ -251,3 +252,132 @@ void single_qubit_control_single_qubit_dense_matrix_gate_simd(
...
+    const CTYPE matrix[4], CTYPE* state, ITYPE dim, UINT inner_qc) {
+    MPIutil& m = MPIutil::get_inst();
+    const UINT rank = m.get_rank();
+    ITYPE dim_work = dim;
+    ITYPE num_work = 0;
+    CTYPE* t = m.get_workarea(&dim_work, &num_work);
+    assert(num_work > 0);
+    CTYPE* si = state;
+
+    if (target_qubit_index < inner_qc) {  // control, target: inner, inner
```
KowerKoint, line 275: small fix: This comment should be written in one line below.

src/csim/update_ops_control_single_target_single.cpp
```diff
@@ -251,3 +252,132 @@ void single_qubit_control_single_qubit_dense_matrix_gate_simd(
...
+                control_qubit_index, control_value, target_qubit_index, matrix,
+                state, dim);
+        } else {  // control, target: outer, inner
+            const UINT control_rank_bit = 1 << (control_qubit_index - inner_qc);
+            if (((rank & control_rank_bit) && (control_value == 1)) ||
+                (!(rank & control_rank_bit) && (control_value == 0)))
+                single_qubit_dense_matrix_gate(
+                    target_qubit_index, matrix, state, dim);
+        }
+    } else {  // control, target: inner, outer
```
KowerKoint, line 287: small fix: This comment should be written in 6 lines below.

src/csim/update_ops_matrix_dense_multi.cpp
KowerKoint: I missed that. That's right.

src/csim/stat_ops_expectation_value.cpp
```diff
@@ -260,15 +265,125 @@ double expectation_value_multi_qubit_Pauli_operator_partial_list_single_thread(
...
+#endif
+                for (j = 0; j < dim_work; ++j) {
+                    ITYPE basis_1 = state_index + j + (pair_rank << inner_qc);
+                    ITYPE basis_0 = basis_1 ^ bit_flip_mask;
+                    UINT sign_0 =
+                        count_population(basis_0 & phase_flip_mask) % 2;
+
+                    sum += _creal(
+                        state[basis_0 & inner_mask] *
+                        conj(recvptr[basis_1 & (dim_work - 1)]) *
```
KowerKoint, line 329: I think it is not guaranteed that basis_0 and basis_1 are in same `i`(workspace-index).

ex)
If
outer_qc=1
inner_qc=3
num_work=1<<1
bit_flip_mask=0b1101
,

an opponent of index 0b0000 (mpi_rank=0b0, i=0b0, j=0b00) is index 0b1101 (mpi_rank=0b1,i=0b1,j=0b01).

m-ymzk: This double loop of ```i``` and ```j``` is based on the sending side: the loop of ```i``` sweeps the communicating location in dim_work units.
Then, the loop of ```j```, which operates on the receiving side, scans the received chunks in order on the basis_1 side and obtains the corresponding basis_0 by exclusive-OR with the bit_flip_mask.

In other words, the sender accesses the state vector in order, while the receiver moves back and forth across its own state vector, ignoring the value of ```i```.

KowerKoint: Sorry. I was wrong. Thank you for explaining.

README_MPI.md
```diff
@@ -0,0 +1,341 @@
...
+
+    - state.to_string()
+      Each rank outputs only state information that has itself.
+        ```
+        to_string() example
+        -- Output from rank 0 process ------------------
+         *** Quantum State ***
+         * Qubit Count : 10
+         * Local Qubit Count : 9
+         * Local Qubit Count : 1       // MPI-size=2
```
KowerKoint, line 39:
```suggestion
         * Global Qubit Count : 1       // MPI-size=2
```

KowerKoint (approved):
Thank you for implementing and refactoring!

README_MPI.md
m-ymzk: Thanks for the review, I corrected it.

src/csim/update_ops_control_single_target_single.cpp
m-ymzk: Thanks, I corrected it.

src/csim/update_ops_control_single_target_single.cpp
m-ymzk: Thanks, I corrected it.

forest1040 (approved):
👍 Thank you!!!

PR #454, opened 2022-12-20
I fix submenu to corrected the columns.

forest1040:
`Quantum circuit` page title had three #.
It was have to two #.
And `Construct the quantum circuit` title had for #.
I changed three #.


Hiroya-W (approved):
Thank you! LGTM!

PR #459, opened 2023-01-05
parse exponential notation in QASM -> Qulacs convertion

kotamanegi:
In QASM -> Qulacs convertion function, previous code only accepts "normal" decimal value like 0.8, 0.023 etc.
I expanded it to accept exponential notation like 8.0e-1, 2.3e-2 etc.

I also used formatter to match with Qulacs' python coding standards.


forest1040 (approved):
👍

KowerKoint (approved):
It can't match `.5`, but I think it's OK.

kotamanegi:
> It can't match .5, but I think it's OK.
It might be a pitfall for this feature.
I added CAVEAT comment on code.

PR #460, opened 2023-01-06
Build wheel for arm64 macOS correctly

### LINKED ISSUE #431: Cannot load shared library on M1 macOS
On M1 macOS, we cannot load qulacs shared library to run Python code.
Here is the minimal reproduction code:
```python
```
And got the following error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/xxx/.pyenv/versions/3.10.8/lib/python3.10/site-packages/qulacs/__init__.py", line 1, in <module>
    from qulacs_core import *
ImportError: dlopen(/Users/xxx/.pyenv/versions/3.10.8/lib/python3.10/site-packages/qulacs_core.cpython-310-darwin.so, 0x0002): tried: '/Users/xxx/.pyenv/versions/3.10.8/lib/python3.10/site-packages/qulacs_core.cpython-310-darwin.so' (mach-o file, but is an incompatible architecture (have (x86_64), need (arm64e)))
```

ikanago:
close #431 

Major changes:
- Use Python 3.10 on CI to use cached interpreter on GitHub Actions runner
- Upload wheels to GitHub Artifact to test wheels locally
- Set `-mtune=native -march=native` only for gcc; clang does not accept these options


forest1040 (approved):
:+1:  Thank You!

KowerKoint (approved):
LGTM

kotamanegi (approved):
:+1:

PR #478, opened 2023-01-17
Fix CMake warnings

### LINKED ISSUE #406: Remove cmake warning
We have warnings while building:
```
Compatibility with CMake < 2.8.12 will be removed from a future version of CMake.

Update the VERSION argument <min> value or use a ...<max> suffix to tell CMake that the project does not need compatibility with older versions.
```
This is not harmful, but is annoying to test cmake variables.
So upgrade cmake requirement to remove this warning.

Hiroya-W:
- close #406 

Fixed the following CMake warning.

-  Fix CMake Deprecation Warning in GoogleTest source
	- Update GoogleTest to latest version
- Use CMP0135 policy's new behavior

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,7 +150,7 @@ if(USE_TEST)
 	FetchContent_Declare(
 		googletest_fetch
 		GIT_REPOSITORY https://github.com/google/googletest
-		GIT_TAG release-1.8.1
+		GIT_TAG release-1.12.1
 	)
 	FetchContent_GetProperties(googletest_fetch)
 	if(NOT googletest_fetch_POPULATED)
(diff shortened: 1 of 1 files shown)


forest1040 (approved):
:+1: Thank you!

KowerKoint (approved):
:+1:
GIT_TAG of googletest in `cmake_script/FetchContent.cmake`'s rST should also be changed.

Hiroya-W:
> 👍 GIT_TAG of googletest in `cmake_script/FetchContent.cmake`'s rST (line 32, 98) should also be changed.

@KowerKoint 
Thank you. I had missed it. 
But the one in `cmake_script/FetchContent.cmake` seems to be an example. I don't think it needs to be changed.

ikanago:
I've checked that CMake warnings has gone away.

ikanago (approved):
I've checked that CMake warnings has gone away. 👍 
I agree with @Hiroya-W 's opinion.

PR #479, opened 2023-01-17
Should not use `lambda` as an argument name

### LINKED ISSUE #477: Should not use `lambda` as an argument name
For example, `U1` function has the following signature

```py
>>> from qulacs.gate import U1
>>> help(U1)
```

```
Help on built-in function U1 in module qulacs_core.gate:

U1(...) method of builtins.PyCapsule instance
    U1(index: int, lambda: float) -> qulacs_core.QuantumGateMatrix
    
    Create QASM U1 gate
```

However, `lambda` arguments cannot be used as keyword arguments. A syntax error will occur.

```py
>>> from qulacs.gate import U1
>>> U1(0, 0.2)
<qulacs_core.QuantumGateMatrix object at 0xffff8b60dab0>
>>>
>>> U1(0, lambda=0.2)
  File "<stdin>", line 1
    U1(0, lambda=0.2)
                ^
SyntaxError: invalid syntax
>>> 
```

In the generation of stub files for qulacs, there is a workaround due to a similar problem.

ref: https://github.com/Qulacs-Osaka/qulacs-osaka/pull/224

Hiroya-W:
- close #477 

Because `lambda` cannot be used as a keyword argument, this change has no effect to users.

- Add `_` as a suffix
	- change `labmda` to `lambda_`.


forest1040 (approved):
:+1:

KowerKoint (approved):
:+1:

kotamanegi (approved):
:+1:

Issue #481, opened 2023-01-23
Fail to build on aarch64 linux


r-imai-quantum:
gcc in aarch64 doesn't have `-mfpmath` option. It looks like it should be omitted if the build process is for aarch64.
https://github.com/qulacs/qulacs/blob/a9e6d82de12dd5d11cdfa56d4a53317c0660c997/CMakeLists.txt#L41

PR #482, opened 2023-01-23
Enable mfpmath option only if x86

r-imai-quantum:
related issue: https://github.com/qulacs/qulacs/issues/481

- Updated CMakeLists.txt
- I ran `pip install .[dev]` in my chromebook (aarch64) then I saw the installation can be done without errors.


KowerKoint (commented):
Thank you. I confirm the issue and the solution by using [amd64v8:ubuntu](https://hub.docker.com/r/arm64v8/ubuntu/) image run on Apple Silicon.

I wonder if the problem also check `"AMD64"` like line 20, but I found that `${CMAKE_HOST_SYSTEM_PROCESSOR}` (which is got from `uname -m`) cannot be the value `"AMD64"`, so it's OK.
ref: https://stackoverflow.com/questions/45125516/possible-values-for-uname-m

KowerKoint (approved):

forest1040 (approved):
The AMD64 is integrated with the x86-64, so this looks fine.
I thought it might be AMD64 if I used MSVC as a compiler, but after all Windows builds with gcc -8, so I think that's fine.

ikanago (approved):
I have checked `mfpmath=both` is disabled on aarch64 and the build finished without problem.  Thank you for your fix!

r-imai-quantum:
Thank you all for your comments and approval!

PR #523, opened 2023-03-01
fix GPU

forest1040:
A recent fix caused a build error in the GPU version, so I fixed it.


ikanago (approved):
Thank you for fixing errors I made...!
I will create cmake target that builds test without running in another issue.

KowerKoint (approved):
:+1:

PR #551, opened 2023-05-11
Splitted python tests using pytest

### LINKED ISSUE #424: split python test files
pythontestが一つのファイルに全部詰め込まれていて見にくいので、複数のファイルに分割したい。

Hiroya-W:
- Close #424 

## Overview

Instead of Python unittest, I used pytest to split the test file.
By using pytest, it is easier to see the results of Python test runs and the causes of errors.


ikanago (approved):
Thank you for hard work to split and tidy Python tests!
I've checked that all the tests are preserved through this revision by some `grep`s.

ikanago (changes requested):
Wheel CI on Ubuntu fails because the previous test command is specified.
Could you fix to use the newer one?
https://github.com/qulacs/qulacs/actions/runs/4955270756

KowerKoint (approved):
LGTM except `test-command` in `pyproject.toml`, which is indicated by ikanago.

herring101 (approved):
LGTM!

Hiroya-W:
@ikanago Thank you! Good suggestion. I have fixed it.

ikanago (approved):
👍 💯

PR #552, opened 2023-05-16
Calculate a more accurate value by increasing the number of samples

Hiroya-W:
## Overview

Some tests in `NoisyEvolutionTest_fast` sometimes fail.
This PR will be fixed to calculate more accurate values by increasing the number of samples. However, I'm not sure if this number of samples is sufficient. In particular, `NoisyEvolutionTest_fast.tekitouu` seem to have a large variation in results. (So, I set a larger number of samples than others)
We need to monitor the tests even after we merge this PR.


KowerKoint (approved):
OK. Let's observe if this is enough for a while.

ikanago (approved):
👍

forest1040 (approved):
👍

PR #557, opened 2023-05-23
Rewrite get matrix function in cpp from py

gandalfr-KY:
The get_matrix function has been added to the general_quantum_operator class method.
It was returning an error when qubit was 1, which has been fixed.


KowerKoint (commented):
Thank you for nice implementation!
Could you add some tests?

gandalfr-KY:
Added a test. There was also a problem with zero elements remaining in the calculation results, which has been corrected.

KowerKoint (approved):
Thank you!

forest1040 (approved):
👍

PR #564, opened 2023-06-15
Collect missing coverage

### LINKED ISSUE #555: Some coverage is missing
For example, look at Codecov, `QuantumGate_Probabilistic` class constructor is not covered.

https://app.codecov.io/gh/qulacs/qulacs/blob/main/src/cppsim/gate_general.hpp#L35

However, it is actually used in the test.

https://github.com/qulacs/qulacs/blob/efa48d3d21300980df8bc21371eb098229e914a6/test/cppsim/test_state_dm.cpp#L61-L84

This is because we forgot to collect the `*.gcda` of the test programs we ran.

https://github.com/qulacs/qulacs/blob/efa48d3d21300980df8bc21371eb098229e914a6/CMakeLists.txt#L459

Hiroya-W:
- #555 

## Overview

Collect missing coverage. However, this also collects coverage of the test program itself. 
As a solution, I tried to separate classes and functions implemented in `*.hpp` into `*.cpp`. I have succeeded in doing this, but because of significant source code changes, I can't guarantee that there will be no bugs.

In this case, it is better to allow the test program itself to be collected...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -456,7 +456,9 @@ if(USE_TEST)
 		add_custom_target(coverage
 			DEPENDS test
 			COMMAND mkdir -p "${CMAKE_CURRENT_SOURCE_DIR}/coverage"
-			COMMAND lcov -d "${CMAKE_CURRENT_SOURCE_DIR}/build/src/csim/CMakeFiles/csim_static.dir" -d "${CMAKE_CURRENT_SOURCE_DIR}/build/src/cppsim/CMakeFiles/cppsim_static.dir" -d "${CMAKE_CURRENT_SOURCE_DIR}/build/src/vqcsim/CMakeFiles/vqcsim_static.dir" -c -o "${CMAKE_CURRENT_SOURCE_DIR}/coverage/coverage.info"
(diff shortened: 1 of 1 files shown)


ikanago (approved):
Including test code into coverage target is little weird, but it is acceptable to collect coverage properly...
Roughly, how much test code is needed to split everything in .cpp to .hpp?

Hiroya-W:
https://github.com/qulacs/qulacs/commit/39b304635553dd433ed68cb94944070225bdf77b is the commit with the file split.
Looking at the current coverage information (https://app.codecov.io/gh/qulacs/qulacs/tree/555-collect-missing-coverage/src%2Fcppsim), the Missed line in vqcsim is 164 lines (optimizer.hpp, parametric_gate.hpp, parametric_gate_factory.hpp), so writing tests is easy.
However, cppsim has 1762 missing lines (gate_general.hpp, state.hpp, state_dm.hpp, or some functions), so we need to consider testing under some conditions (GPU, MPI), performance, etc.

KowerKoint (approved):
OK.
That'll be so big change, but it's necessary.

PR #565, opened 2023-06-15
remove std::find(), precalc where parametric gates exist

KowerKoint:
`std::find()` conducts linear search every time, which is slow.
This PR resolves it by getting which gate is parametric previously.


ikanago (approved):
More efficiently and more readable! LGTM

herring101 (approved):
👍

PR #559, opened 2023-05-25
Pass codecov check on GitHub Actions

Hiroya-W:
## Overview

Added Codecov config file (`informational: true`) in order for Codecov checks on GitHub Actions to always pass.
Coverage threshold checks, etc., I think can be set up after the test has been sufficiently written.

PR #568, opened 2023-06-20
Pass codecov patch check

Hiroya-W:
## Overview

Add configs in order for Codecov checks to always pass as well as project check( #559 ).


ikanago (approved):
🙆‍♂️

PR #576, opened 2023-07-20
Add missing test cases in circuit test

Hiroya-W:
## Overview

Add test for `add_*_gate` method to add gate in QunatumCircuit.
Because there are tests where gate has the correct matrix, I now check if the add operations to the circuit can be used correctly.

### DIFF
--- a/test/cppsim/test_circuit.cpp
+++ b/test/cppsim/test_circuit.cpp
@@ -135,7 +135,9 @@ TEST(CircuitTest, CircuitBasic) {
 
     target = random.int32() % n;
     angle = random.uniform() * 3.14159;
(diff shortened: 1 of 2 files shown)


m-ymzk:
I would appreciate it if you could fix the code for multi-cpu.
https://github.com/qulacs/qulacs/blob/d6306cecc754f315e87cd4b2ee53d592528ae28c/test/cppsim/test_circuit_multicpu.cpp#L136-L138

Hiroya-W:
> I would appreciate it if you could fix the code for multi-cpu as well.

Thank you for suggestion. I added testcases for multicpu.

forest1040 (approved):
Thx! :+1:

PR #581, opened 2023-08-29
Fix diagonal matrix gate for MPI

Akihiro-Tabuchi:
This PR fixes diagonal matrix gate for MPI.

There are diagonal matrix gate update functions for MPI, but they are not called. This PR modifies these functions to be called properly and add a test.


forest1040 (approved):
thank you! :+1:

PR #585, opened 2023-09-05
Ubuntu 22.04 LTS installation instructions

WrathfulSpatula:
Hi, I'm Dan Strano from over at https://github.com/unitaryfund/qrack. I build and install your library regularly, and it took me several hours today to figure out how to install on my Ubuntu 22.04 LTS (and 20.04 LTS) systems. However, it's very simple to communicate a workaround (on close-to-stock LTS). If you like this documentation change, I think it broadens your installation support significantly.

(Thanks for the open-source simulator library!) :v:

PR #592, opened 2023-09-14
add cmake to build_system.requires

KowerKoint:
related to #585 
In the environemnt which `cmake` executable is replaced to python-wrapper version, pip's install-dev could not find `cmake` python package.
To resolve this, I added `cmake` to `build_system.requires`.


KowerKoint:
There seems to be no way to avoid insalling `cmake` executable by `build-system.requires` installation.

PR #598, opened 2023-10-31
Adding swap insertion feature to circuit optimizer

Akihiro-Tabuchi:
This PR adds a swap/fused-swap gate insertion feature to circuit optimizers for multi-cpu execution.

As described in README_MPI.md, you can use this feature by adding `swap_level` argument to optimize() or optimize_light() method.

Example code:
```python
from qulacs import QuantumCircuit
from qulacs.circuit import QuantumCircuitOptimizer

state = QuantumState(nqubits, True)
circuit = QuantumCircuit(nqubits)
# add some gates to the circuit here

qco = QuantumCircuitOptimizer()
qco.optimize(circuit, blocksize, swap_level=2)

circuit.update_quantum_state(state)
```

The value for `swap_level` means:
- 0: No SWAP/FusedSWAP gate insertion (default)
- 1: Insert SWAP/FusedSWAP gates to reduce communication without changing gate order
- 2: Insert SWAP/FusedSWAP gates to reduce communication with changing gate order

The number of processes is automatically obtained from MPI environment at runtime.
To optimize the circuit for a specific number of processors, add the `mpi_size` argument to the QuantumCircuitOptimizer constructor.
```python
qco = QuantumCircuitOptimizer(mpi_size=4)
qco.optimize(circuit, blocksize, swap_level=1) # the circuit is optimized for 4 processors
```

The CIs are passed except Ubuntu CI (GCC10 build with -fsanitizer=address enabled). This CI has also failed in other pull requests, so it is unlikely to be relevant for this pull request.

### DIFF
--- a/python/cppsim_wrapper.cpp
+++ b/python/cppsim_wrapper.cpp
@@ -1525,12 +1525,13 @@ Matrix Representation
         "from json string", py::return_value_policy::take_ownership);
 
     py::class_<QuantumCircuitOptimizer>(mcircuit, "QuantumCircuitOptimizer")
-        .def(py::init<>(), "Constructor")
+        .def(py::init<UINT>(), "Constructor", py::arg("mpi_size") = 0)
         .def("optimize", &QuantumCircuitOptimizer::optimize,
             "Optimize quantum circuit", py::arg("circuit"),
-            py::arg("block_size"))
+            py::arg("block_size"), py::arg("swap_level") = 0)
         .def("optimize_light", &QuantumCircuitOptimizer::optimize_light,
-            "Optimize quantum circuit with light method", py::arg("circuit"))
+            "Optimize quantum circuit with light method", py::arg("circuit"),
+            py::arg("swap_level") = 0)
         .def("merge_all", &QuantumCircuitOptimizer::merge_all,
             py::return_value_policy::take_ownership, py::arg("circuit"));
 
--- a/src/cppsim/circuit_optimizer.cpp
+++ b/src/cppsim/circuit_optimizer.cpp
@@ -457,7 +457,7 @@ bool QuantumCircuitOptimizer::can_merge_with_swap_insertion(
 }
 
 bool QuantumCircuitOptimizer::needs_communication(
-    const UINT gate_index, QubitTable& qt) {
+    const UINT gate_index, const QubitTable& qt) {
     auto gate = circuit->gate_list[gate_index];
     auto logical_qubits = get_qubits_needing_communication(gate);
     return std::any_of(
@@ -468,8 +468,9 @@ bool QuantumCircuitOptimizer::needs_communication(
 }
 
 UINT QuantumCircuitOptimizer::move_gates_without_communication(
-    const UINT gate_idx, QubitTable& qt,
-    std::multimap<const QuantumGateBase*, const QuantumGateBase*>& dep_map,
+    const UINT gate_idx, const QubitTable& qt,
+    const std::multimap<const QuantumGateBase*, const QuantumGateBase*>&
+        dep_map,
     std::unordered_set<const QuantumGateBase*>& processed_gates) {
     const UINT num_gates = circuit->gate_list.size();
     UINT moved_gates = 0;
@@ -559,7 +560,7 @@ UINT QuantumCircuitOptimizer::move_matching_qubits_to_local_upper(
 }
 
 UINT QuantumCircuitOptimizer::rearrange_qubits(const UINT gate_idx,
-    std::unordered_set<UINT> next_local_qubit, QubitTable& qt) {
+    const std::unordered_set<UINT>& next_local_qubit, QubitTable& qt) {
     UINT num_inserted_gates = 0;
 
     std::unordered_set<UINT> cur_local_qubits(
@@ -614,7 +615,8 @@ UINT QuantumCircuitOptimizer::rearrange_qubits(const UINT gate_idx,
     // TODO: if fast qubit swapping between global qubits is implemented,
     //       global qubits should be rearranged to be contiguous.
     UINT total_swap_width = import_qubits.size();
-    int extra_local_qc = exportable_qubits.size() - import_qubits.size();
+    int extra_local_qc =
+        (int)exportable_qubits.size() - (int)import_qubits.size();
 
     for (UINT i = 0; i < swap_global_idx_list.size(); i++) {
         for (UINT j = i + 1; j < swap_global_idx_list.size();) {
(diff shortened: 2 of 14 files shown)


src/cppsim/circuit_optimizer.cpp
```diff
@@ -198,3 +235,680 @@ QuantumGateMatrix* QuantumCircuitOptimizer::merge_all(
...
+        }
+    }
+
+    // fuse swaps
+    // e.g. swap_global_idx_list = [2 6] -> [2]
+    //      swap_width_list      = [1 1] -> [3]
+    // TODO: if fast qubit swapping between global qubits is implemented,
+    //       global qubits should be rearranged to be contiguous.
+    UINT total_swap_width = import_qubits.size();
+    int extra_local_qc = exportable_qubits.size() - import_qubits.size();
```
KowerKoint, line 617: All supported compilers express signed integers as Two's Complement so this works, but it is forced from C++20.
`(int)exportable_qubits.size() - (int)import_qubits.size()` is safer.

src/cppsim/circuit_optimizer.cpp
```diff
@@ -198,3 +235,680 @@ QuantumGateMatrix* QuantumCircuitOptimizer::merge_all(
...
+        is_global_qubit);
+    std::copy_if(controls_gate2.cbegin(), controls_gate2.cend(),
+        std::inserter(global_control_gate2, global_control_gate2.begin()),
+        is_global_qubit);
+
+    return global_control_gate1 == global_control_gate2;
+}
+
+bool QuantumCircuitOptimizer::needs_communication(
+    const UINT gate_index, QubitTable& qt) {
```
KowerKoint, line 460:
```suggestion
    const UINT gate_index, const QubitTable& qt) {
```

src/cppsim/circuit_optimizer.cpp
```diff
@@ -198,3 +235,680 @@ QuantumGateMatrix* QuantumCircuitOptimizer::merge_all(
...
+    return std::any_of(
+        logical_qubits.cbegin(), logical_qubits.cend(), [&](auto& l_q) {
+            // true if physical qubit is global qubit
+            return qt.l2p[l_q] >= local_qc;
+        });
+}
+
+UINT QuantumCircuitOptimizer::move_gates_without_communication(
+    const UINT gate_idx, QubitTable& qt,
+    std::multimap<const QuantumGateBase*, const QuantumGateBase*>& dep_map,
```
KowerKoint, line 472:
```suggestion
    const UINT gate_idx, const QubitTable& qt,
    const std::multimap<const QuantumGateBase*, const QuantumGateBase*>& dep_map,
```

src/cppsim/circuit_optimizer.cpp
```diff
@@ -198,3 +235,680 @@ QuantumGateMatrix* QuantumCircuitOptimizer::merge_all(
...
+                "QuantumCircuitOptimizer::move_matching_qubits_to_local_upper()"
+                " "
+                "no enougth matched qubits");
+        }
+    }
+    return num_inserted_gates;
+}
+
+UINT QuantumCircuitOptimizer::rearrange_qubits(const UINT gate_idx,
+    std::unordered_set<UINT> next_local_qubit, QubitTable& qt) {
```
KowerKoint, line 562:
```suggestion
    const std::unordered_set<UINT>& next_local_qubit, QubitTable& qt) {
```

KowerKoint (approved):
Thank you! Seems good.
I have a few small fix suggestions.

Akihiro-Tabuchi:
@KowerKoint 
Thank you for your review and suggestions.
I applied your all suggestions by c5776b8

KowerKoint:
Thank you.
Currently CI of  readthedocs  is failing, so I'll fix it and merge this.

python/cppsim_wrapper.cpp
```diff
@@ -1525,12 +1525,13 @@ Matrix Representation
         "from json string", py::return_value_policy::take_ownership);
 
     py::class_<QuantumCircuitOptimizer>(mcircuit, "QuantumCircuitOptimizer")
-        .def(py::init<>(), "Constructor")
+        .def(py::init<UINT>(), "Constructor", py::arg("mpi_size") = 0)
         .def("optimize", &QuantumCircuitOptimizer::optimize,
```
m-ymzk, line 1529: For backwards compatibility, shouldn't this be handled by an addition rather than a change to the constructor?
If the argument mpi_size is not specified, we would like it to behave basically the same as before.
The same applies to optimize_light.
The same also applies to the API as visible from C++.

m-ymzk (changes requested):

m-ymzk:
Sorry for the very late reply.
I would like to ask you to revisit the modified part of the API and if possible also modify benchmark/bench_circuit.py.

python/cppsim_wrapper.cpp
```diff
@@ -1525,12 +1525,13 @@ Matrix Representation
         "from json string", py::return_value_policy::take_ownership);
 
     py::class_<QuantumCircuitOptimizer>(mcircuit, "QuantumCircuitOptimizer")
-        .def(py::init<>(), "Constructor")
+        .def(py::init<UINT>(), "Constructor", py::arg("mpi_size") = 0)
```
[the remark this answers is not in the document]
Akihiro-Tabuchi: I think there's a misunderstanding.
`mpi_size` is an optional argument of  the constructor, so it is obtained internally if not specified.
This argument is provided for circuit and optimizer debugging purposes and is not intended for normal use.

In test code, the optimizer works correctly without 'mpi_size' argument actually.
https://github.com/Akihiro-Tabuchi/qulacs/blob/c5776b84f27510d360593b18929fa474c8b552d6/python/tests/multi_cpu/test_circuit_optimizer.py#L64

Of course, the optimizer also works correctly with 'mpi_size' argument.
https://github.com/Akihiro-Tabuchi/qulacs/blob/c5776b84f27510d360593b18929fa474c8b552d6/python/tests/multi_cpu/test_circuit_optimizer.py#L83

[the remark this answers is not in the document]
m-ymzk: Sorry, I was mistaken.
I did indeed confirm that the parameters added by this modification are optional.

python/cppsim_wrapper.cpp
m-ymzk: Sorry, I was mistaken.
I did indeed confirm that the parameters added by this modification are optional.

m-ymzk (approved):

KowerKoint:
Thank you.
Yes. The failure of `StateTest.SamplingSuperpositionState` is not related to this change.

PR #601, opened 2023-11-13
upgrade doc dependencies

### LINKED ISSUE #600: fix readthedocs CI Error
https://stackoverflow.com/questions/77257145/sphinx-autoapi-error-module-object-has-no-attribute-doc-with-various-sphi
upgrading sphinx and sphinx-auto-api seems to solve.

KowerKoint:
close #600
We have to upgrade sphinx-api (ref: https://stackoverflow.com/questions/77257145/sphinx-autoapi-error-module-object-has-no-attribute-doc-with-various-sphi ).
I tried to explicitly new versions of sphinx, sphinx-api and others on `pyproject.toml`, but sphinx dep version of breathe, exhale and myst-parser is very complex so I remove version restrictions and let pip automatically resolve conflict.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,15 +62,15 @@ ci = [
 ]
 
 doc = [
-    "sphinx == 4.5.0",
-    "sphinx-rtd-theme == 1.0.*",
-    "breathe == 4.33.*",
-    "exhale == 0.3.*",
-    "nbsphinx == 0.8.*",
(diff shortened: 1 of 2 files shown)


KowerKoint:
exhale irequires too low sphinx version and causes version conflict.
so use unmerged GitHub branch for workaround.

forest1040 (approved):
👍

PR #620, opened 2024-03-19
ignore stub

### LINKED ISSUE #619: Not track stub files by Git
The result of pybind11-stubgen differs between python3.10 and python3.11, so `*.pyi` files generated by `update_stubs.sh` depends on the environment.
This issue will include `*.pyi` into `.gitignore` and stop tracking stub files.
Developers can generate stub files by manually running `update_stub.sh`.
GitHub Actions to build and deploy document and wheel generates stub files by the runners.

KowerKoint:
close #619 
The format of stubs file depend on the version of pybind11-stubgen, so tracking them is inappropriate.
In wheel build, this workflow installs qulacs once, then generates stubs by pybind11-stubgen, and finally builds qulacs package again with including stubs.


KowerKoint:
I still think this "install -> stubgen -> reinstall" flow is weired, but pybind11 does not supply stubgen and other stubgen-related programs require installed-packages.
To avoid this, we can

- always type `.pyi` files manually, then track them on Git and test them by CI
- stop directly using pybind11 API in `cppsim_wrapper.cpp` and develop a wrapper to do both of follows:
  - generate `pyi` files by reading the type-signatures on C++
  - pass pybind11 and build python module

KowerKoint:
I find the setuptools-like project 
named robotpy-build ref: https://github.com/pybind/pybind11/issues/2350#issuecomment-1126255196 which seems to generate stub files when the package is installed.

By the way, the next release of nanobind will support stub generation officialy (ref: https://nanobind.readthedocs.io/en/latest/changelog.html#version-2-0-0-tba ).
I'll try this on scaluq first.

ikanago:
There was no Wheel CI run in this branch ever, so I triggered it by workflow_dispatch.
Then I got errors on all the build matrices. I'm investigating on this.
https://github.com/qulacs/qulacs/actions/runs/8337349995/job/22815950272

.github/workflows/wheel.yml
```diff
@@ -41,6 +40,15 @@ jobs:
...
           - os-arch: "macosx_arm64"
             os: "macos-12"
+          - cibw-python: "cp38"
+            python-version: "3.8"
+          - cibw-python: "cp39"
+            python-version: "3.9"
+          - cibw-python: "cp310"
+            python-version: "3.10"
+          - cibw-python: "cp311"
+            python-version: "3.11"
```
ikanago, line 50: You forget cp37 and Python 3.7 pair.
Or we can drop the support for Python 3.7 as its security support ended 8 month ago.
https://endoflife.date/pythong

.github/workflows/wheel.yml
```diff
@@ -61,13 +69,38 @@ jobs:
       - name: Install Python dependencies
         run: python -m pip install cibuildwheel twine
 
+      - name: Install boost if ubuntu
+        if: ${{ contains(matrix.os, 'ubuntu') }}
+        run: sudo apt update && sudo apt install libboost-dev
```
ikanago, line 74: If we do not run `apt upgrade`, we can remove `apt update`, right?

ikanago (changes requested):

.github/workflows/wheel.yml
KowerKoint: OK.
I'll remove Python 3.7.

.github/workflows/wheel.yml
KowerKoint: Really? I think `libboost-dev` may not found in the repository without `apt update`, but actually it is always found on GitHub Actions Runner.

KowerKoint:
I removed Python 3.7 wheel build, and added Python 3.12 wheel build.

ikanago (approved):
LGTM!  Thank you!

Issue #618, opened 2024-03-14
Flexible Block size which depends on GPU


KowerKoint:
in gpusim, we use fixed block-size (ex: `1024` for most of single-qubit and double-qubit gate-apply, `512` for triple-qubit and quad-qubit dense-matrix-gate-aplly).

In most of GPU, the fixed-value efficiently works, but it sometimes failes.
For example, `block=512` in `quad_qubit_dense_matrix_gate_gpu` doesn't work for H100.
This is probably because H100 can assign many registers per one thread and registers in a block cannot afford with `block=512` when assigning maximum number of registers.

We can get appropriate (maximizing potential occupancy) block-size dynamically using `cudaOccupancyMaxPotentialBlockSize`.

KowerKoint:
Benchmark by 4xA100

All bench was run with n = 28
| function| current block_size | current exec time | recommended block_size | changed exec time |
|--|--|--|--|--|
| initialize_quantum_state| 1024 | 28ms | 1024 | 28ms |
| initialize_Haar_random_state_with_seed | 512 | 2614ms | 768 | 2630ms |
| state_norm_squared (1<<5 loops per threads) | 256 | 814ms | 1024 | 835ms |

KowerKoint:
I haven't check `multi_qubit_control_multi_qubit_dense_matrix_gate` and the functions which are not current used from cppsim, but I found that using `cudaOccupancyMaxActivateBlockSize` will not lose performance.

KowerKoint:
In addition, I found that `loop_dim` of reduction functions (ex: `get_squared_norm`) is not set to good value when `n` is big.
CUDA's `attomicAdd` is very slow when grid size is high.

I'll do the experiment to found a good `loop_dim` value of `get_squared_norm`.

KowerKoint:
[A 100]

|n qubits|execution time when loop_dim=dim>>0|dim>>1|dim>>2|dim>>3|dim>>4|dim>>5|dim>>6|dim>>7|dim>>8|dim>>9|dim>>10|dim>>11|dim>>12|dim>>13|dim>>14|
|--|--|--|--|--|--|--|--|--|--|--|--|--|--|
|16|1.88ms|0.67ms|0.30ms|0.16ms|0.10ms|0.08ms|**0.07ms**|0.08ms|0.12ms|0.20ms|0.37ms|0.71ms|1.37ms|2.69ms|4.64ms|
[...]

PR #626, opened 2024-03-21
Haar_random_state GPU bug

### LINKED ISSUE #625: A bug of QuantumStateGpu::set_Haar_random_state when nqubits >= 19
In https://github.com/qulacs/qulacs/blob/98a2751666dcdcc787df5db7e6e739d77a6ff1f8/src/gpusim/memory_ops.cu#L147 , maximum blocksize x gridsize = 512 x 512 = 2^18.
So, state[2**18:] is not initialized correctly.

```py
from qulacs import QuantumStateGpu

state = QuantumStateGpu(19)
state.set_Haar_random_state()
v = state.get_vector()
print(v[0])
print(v[1])
print(v[1 << 17])
print(v[1 << 18])
print(v[(1 << 18) + 1])
print(v[-1])
```

```
(-0.0004327631602487252-0.0007006227218964226j)
(-0.0008768382449739581+0.00012370432683280022j)
(-0.00045247259213412145+0.001093584498417163j)
0j
0j
0j
```

KowerKoint:
close #625  
I don't know why old dev insert `min(, 512)`, but this works with nqubits=25 in my environment.

```py
from qulacs import QuantumStateGpu

state = QuantumStateGpu(25)
state.set_Haar_random_state()
print(state.get_marginal_probability([1] * 25))

```

### DIFF
--- a/src/gpusim/memory_ops.cu
+++ b/src/gpusim/memory_ops.cu
@@ -2,7 +2,7 @@
 
 #include "cuda_runtime.h"
 #include "device_launch_parameters.h"
-//#include <cuda.h>
+// #include <cuda.h>
 #include <assert.h>
 #include <cuComplex.h>
 #include <curand.h>
@@ -144,7 +144,7 @@ __host__ void initialize_Haar_random_state_with_seed_host(void* state,
     // CURAND_RNG_PSEUDO_MT19937 offset cannot be used and need sm_35 or higher.
 
     unsigned int block = dim <= 512 ? dim : 512;
-    unsigned int grid = min((int)(dim / block), 512);
+    unsigned int grid = dim / block;
 
     init_rnd<<<grid, block, 0, *cuda_stream>>>(rnd_state, seed);
     checkCudaErrors(cudaGetLastError(), __FILE__, __LINE__);


KowerKoint:
Tested nqubit=28 with A100

ikanago (approved):
LGTM!

I found the commit that made this change, but i could not figure out why did this change needed...
https://github.com/qulacs/qulacs/commit/6d3a0aa2ce2b7cfce87ed51a602bfc05f122b232
The commit changed the maximum of `block` from 1024 to 512, but the reason is not addressed.

KowerKoint:
In #626, I am thinking of using `cudaOccupancyMaxPotentialBlockSize` to determine block size.

With A100, `cudaOccupancyMaxPotentialBlockSize(min_grid_size, block_size, rand_normal_xorwow)` says `block_size=784` achieves max occupancy.
And I measured the time to run `set_Haar_random_state` with nqubits=28.
The result is as follows:

|block size | time |
|--|--|
| 512 | 2152 ms |
| 784 | 3894ms |
| 1024 | 3881ms |

I found that maximizing occupancy is not always the best and the old developer set a good (but ad-hoc) block size.

ikanago:
Thank you for your investigation!
It can be hard to find this PR for someone who has a question on why `block_size` is 512.
So could you add a summary of your result as a comment? 🙏

KowerKoint:
Sorry, `cudaOccupancyMaxPotentialBlockSize(min_grid_size, block_size, rand_normal_xorwow)` returns `768` (not `784`) with both A100 and H100 machine.
I measured again by repeating 10 times of initializing and calculating an average.
I got the result below.

| block size | A100 time | H100 time |
|--|--|--|
| 512 | 2614 ms | 1379 ms |
| 768 | 2630  ms | 1367 ms |
| 1024 | 2577 ms | 1365 ms |

Result depends on the other effects, but this time I didn't find the difference by block size.
I'll investigate other functions.
If results do not differ, I'll use `cudaOccupancyMaxPotentialBlockSize` because by using that we can get at least one "good" preference which doesn't occur errors.

KowerKoint:
More investigate on #618

PR #631, opened 2024-04-09
Upgrade macOS in Wheel CI

ikanago:
Upgrade macOS version from 12 in Wheel CI.
Since Python 3.8 and 3.9 is not available in macOS 14(arm runner), they are excluded.
And `architecture` option in setup-python is removed. Some options like arm64 does not exist, and [the architecture is selected properly by default.](https://github.com/actions/setup-python/blob/82c7e631bb3cdc910f68e0081d67478d79c6982d/src/setup-python.ts#L98)

### DIFF
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -25,7 +25,7 @@ jobs:
           - compiler: "clang"
             c_compiler: "/usr/local/opt/ccache/libexec/clang"
             cxx_compiler: "/usr/local/opt/ccache/libexec/clang++"
-    runs-on: "macos-12"
+    runs-on: "macos-13"
     env:
       C_COMPILER: ${{ matrix.c_compiler }}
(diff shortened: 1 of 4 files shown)


KowerKoint (approved):
Thank you

KowerKoint:
```
Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```
in the part of uploading artifacts.
I check the file name of `wheel/*.whl` ( https://github.com/qulacs/qulacs/actions/runs/8655428416 ) but they are all different. so I do not know why this is happening.

But downgrading `actions/upload-artifact` to @v3 resolves, it seems to be v4's error.

ikanago:
As of v4, uploading the same **artifact** name (not file name) leads an error.
Maintainer of the action says it is the expected behavior. ref:https://github.com/actions/upload-artifact/issues/478#issuecomment-2040128675
So I named each artifact with Python version, OS and architecture so that they all have different names.

PR #633, opened 2024-04-11
Update stub before documentation build

ikanago:
Python API Reference in the document is empty.
It seems to need stubs to build the API reference, so run `./script/update_stub.sh` before building the document.
Just check .readthedocs.yaml; this PR includes diff pulled from #631 

## Cause
I investigated build logs in readthedocs, and found that [the build for #616](https://qulacs-rtd--616.org.readthedocs.build/en/616/pyRef/qulacs/index.html) was OK and [the one for #620](https://qulacs-rtd--620.org.readthedocs.build/en/620/pyRef/qulacs/index.html) was not.
I looked into the build log for #616, I found the stub files are used to generate API reference.
And the change introduced in #620 was excluding stubs from git source, so the stubs are not available during the documentation build.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,6 +7,9 @@ build:
   apt_packages:
     - cmake
     - libboost-dev
+  jobs:
+    pre_build:
+      - ./script/update_stubs.sh
 
 python:
   install:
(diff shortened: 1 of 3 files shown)


KowerKoint (approved):
Thank you!
I checked the API document is restored in https://qulacs-rtd--633.org.readthedocs.build/en/633/pyRef/qulacs/index.html .

ikanago:
I will merge this after #631 resolves.

Issue #632, opened 2024-04-09
Empty gradients when using `QuantumCircuitOptimizer`


Gopal-Dahale:
## Description
Using `ParametricQuantumCircuit` with `QuantumCircuitOptimizer` (QCO) results in an empty list of gradients as the output of `backprop`. This happens when more than one parametric gate is used in the circuit. I have checked only with `optimize_light` only.

## Expected Behavior
The length and values of the gradients list should be the same with or without the use of QCO.

## Steps to Reproduce
`qulacs v0.6.3` is used. 

### Example code without QCO
```python
from qulacs import Observable, ParametricQuantumCircuit, PauliOperator, QuantumState
from qulacs.circuit import QuantumCircuitOptimizer as QCO

obs = Observable(1)
obs.add_operator(PauliOperator("Z 0", 1.0))

circuit = ParametricQuantumCircuit(1)
circuit.add_parametric_RX_gate(0, 0.2)
circuit.add_parametric_RY_gate(0, 0.3)

state = QuantumState(1)
circuit.update_quantum_state(state)

grads = circuit.backprop(obs)
print(grads)  # output: [-0.18979606097868734, -0.28962947762551555]
```
### With QCO
```python
qco = QCO()
qco.optimize_light(circuit)

state = QuantumState(1)
circuit.update_quantum_state(state)

grads = circuit.backprop(obs)
print(grads) # output: []
```
For this case, printing the circuit information after QCO gives zero parameters.
```
*** Quantum Circuit Info ***
# of qubit: 1
# of step : 1
# of gate : 1
# of 1 qubit gate: 1
Clifford  : no
Gaussian  : no

*** Parameter Info ***
# of parameter: 0
```
### With QCO and a single gate
```python
circuit = ParametricQuantumCircuit(1)
circuit.add_parametric_RX_gate(0, 0.2)

qco = QCO()
qco.optimize_light(circuit)

state = QuantumState(1)
circuit.update_quantum_state(state)

grads = circuit.backprop(obs)
print(grads) # output: [-0.1986693307950612]
```

KowerKoint:
I found the mistake in `optimize_light`.
I'll fix.
Thank you for your report.

ikanago:
solved in #634

PR #634, opened 2024-04-12
do not merge parametric gate in optimize_light

KowerKoint:
#632 
Merging a parametric gate with another gate into DenseMatrix gate will lose its parameter.
In `QuantumCircuit::optimize` do not merge parametric gates.
This pull request make `QuantumCircuit::optimize_light` to do so.


ikanago (approved):
Thank you for your fix!
I've checked that parametric gates are preserved after optimization.
LGTM!

PR #644, opened 2024-05-17
Remove brew upgrade all

KowerKoint:
Resolve error which occurs during upgrating swiftlint.
I think this is responsible to swiftlint, but anyway `brew upgrade` is not needed.


ikanago:
LGTM 👍

ikanago (approved):
LGTM 👍

Issue #645, opened 2024-05-21
`python3.12` installation


BrunoLiegiBastonLiegi:
Hello, installing the cpu version of `qulacs` under `python3.12` simply with `pip install qulacs` fails for me. Separately installing `cmake` and `boost` first, solves the problem though.

ikanago:
Thank you for reporting the issue!
I have confirmed that `pip install qulacs` needs cmake and boost on Python 3.12.
We have introduced the setting to build for Python 3.12 on #620, but we haven't released a new version after that.
We are going to publish a new version including the build for Python 3.12 before long.

scarrazza:
@ikanago thank you for the information. As soon as you release the new version we will be able to merge https://github.com/qiboteam/qibo/pull/1272 and release a new Qibo version including the qulacs backend.

ikanago:
@scarrazza Now qulacs 0.6.4.1 is available. This version is compatible for Python 3.12.
If you have further problems, feel free to inform at this or other issue!

BrunoLiegiBastonLiegi:
Thanks @ikanago, `python 3.12` works like charm now! However, we are having problems with `python 3.9` under `macos`. The reason seems to be that no wheel is available for the ARM version of `macos` with `3.9`, although it's there for all the other python versions. Is there by chance any plan for adding it?

forest1040:
ok! i recheck it. for macos.

KowerKoint:
Python < 3.10 for setup-python was not supported in macos arm64 runner, so we excluded 3.8 and 3.9 in https://github.com/qulacs/qulacs/pull/631 .
Now it seems supported. I'll try to support it.

PR #646, opened 2024-05-24
release-v0.6.4

ikanago:
Not so many new features or fixes, but we should release the package for Python 3.12 as mentioned in #645.


KowerKoint (approved):
:+1:

PR #648, opened 2024-05-30
Fix sphinx-autoapi version to build Python API document properly

ikanago:
Fix sphinx-autoapi on 3.0.0.
Specifying it to "==3.*" leads to install 3.1.1 and fail to build the API document.


forest1040 (approved):
:+1:

PR #657, opened 2024-09-06
do not exclude on python3.8 3.9

KowerKoint:
Python 3.8, 3.9 release of macos arm64 was removed in #631 because actions/setup-python did not support them.
Now they are supported so this PR restore it.


forest1040 (approved):
👍

Issue #659, opened 2024-09-25
Problems with the gradient computation


manuelgpda:
Dear devs,

First of all, thank you for your amazing work.

Recently I found an "issue" that is provided in the attached code snippets. 

Basically, the time taken for gradient computation can drastically change (a factor of 10^6) and even given different results when using the same circuit and the same function depending on how the gate is added. You can in principle add a parametric X gate (option 1), what is slow. Or you can add a multiqubit parametric Pauli rotation specified for a single qubit X. If the latter is done by defining a list `target` with all the qubits in the circuit and placing identities (this is `0`) in the list `pauli_id` (option 2), then the result is correct as expected and it is way faster than the single parametric gate. However, if the lists `target` and `pauli_id` are defined only with qubits on which the rotation is not acting with an identity (option 3), then  the time taken is almost as slow as for a single parametric gate but also if the rotation is made on a qubit that is not the first one, then the result can be wrong (check snippet 2). The rule seems to be that if one includes in both lists all the qubits `j` whose index is smaller than the one on which the rotation is applied, then the multi qubit parametric rotation is faster and gives the right result. I tried to see where the issue is coming from and I believe it is inside the CausalConeSimulator computation.

I believe this might also be related to previously reported issues like [this one](https://discuss.pennylane.ai/t/qualcs-device-and-qml-grad-timings/1717).

Snippet 1
```python
from qulacs import ParametricQuantumCircuit, GradCalculator, Observable
from time import time

n = 2
observable = Observable(n)
observable.add_operator(1.0, "X 0")
observable.add_operator(1.0, "X 1")
observable.add_operator(1.0, "Z 0 Z 1")

theta = 2.2

# option 1
t1 = time()
circuit = ParametricQuantumCircuit(n)
circuit.add_parametric_RX_gate(0, theta)
gcalc = GradCalculator()
result1 = gcalc.calculate_grad(circuit, observable)
t2 = time()
print("Option 1:",result1)
print("Time 1:",t2-t1)

# option 2
t3 = time()
target = [0]
pauli_ids = [1]
circuit = ParametricQuantumCircuit(n)
circuit.add_parametric_multi_Pauli_rotation_gate(target, pauli_ids, theta)
gcalc = GradCalculator()
result2 = gcalc.calculate_grad(circuit, observable)
t4 = time()
print("Option 2:",result2)
print("Time 2:",t4-t3)

# option 3
t5 = time()
target = [0,1]
pauli_ids = [1,0]
[...]

PR #660, opened 2024-09-25
single parametric gates use multi gates under the rug now

manuelgpda:
This is a potential temporary fix to #659 by using `add_parametric_multi_Pauli_rotation_gate` to perform `add_parametric_RX_gate`,` add_parametric_RY_gate`,` add_parametric_RZ_gate` but I did not have the opportunity to test it and there might be still some pieces of code to be removed if this is implemented.


KowerKoint:
The fundamental issue is solved on #661 and #664 . Now this change is not needed, right?

PR #661, opened 2024-10-04
Fix `CausalConeSimulator` by replacing `QuantumGateBase::set_target_qubit_index`

KowerKoint:
close #659 

`QuantumGateBase::set_target_qubit_index` was wrong because it didn't replace target indices of `PauliOperator` on `ClsPauliGate` etc.
I think this function is evil. Replacing target qubit index later makes trouble easily.
So this time I created a new function to make copied gate with targets replaced.

In addition, I found some wrong parts in `CausalConeSimulator` and fix them.

PR #664, opened 2024-10-16
remove useless single-pauli of observable

KowerKoint:
ref: #659 #661 
I found the problem on CausalConeSimulator. on `src/vqcsim/causalcone_simulator.hpp:176`.
All pauli operators on the term has been added, but with qubits which are not in the connected component, `qubit_encode[term_index_list[j]]` is `-1`, which is treated as a large number on `UINT`. Then `PauliOperator::add_single_Pauli` expand `dynamic_bitset` to very large size. This takes so much time and memory.


kotamanegi (approved):
Seems like a good bugfix codes.

PR #673, opened 2025-01-17
specify conf.py explicitly

KowerKoint:
ref: https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/
Now we must specify the path to `conf.py` explicitly.
I found that Japanese version docs https://docs.qulacs.org/ja/latest/index.html is currently English.
This PR makes two `.readthedocs.yml` for English and Japanese docs.
I also configure the path to each `.readthedocs.yml` on the configurations of readthedocs projects.


forest1040 (approved):
👍

forest1040 (approved):
👍

PR #683, opened 2025-05-19
remove Qulacs-Osaka merge information from README.md

### LINKED ISSUE #682: remove Qulacs-Osaka merge information from README.md
Qulacs-Osaka is already merged.

KowerKoint:
close #682


forest1040 (approved):
👍

src/cppsim/general_quantum_operator.cpp
KowerKoint, line 1: MSVC supports only OpenMP 2.0, and `omp declare` directive is not supported.
This parallel algorithm is originally not so good (because every time `SparseComplexMatrixRowMajor::operator+` sees all elements and reconstruct).
I replace that with non-parallel algorithm which create the list of triplets first and finally construct matrix from triplets.

PR #686, opened 2025-06-09
Windows Python3.13 support

KowerKoint:
Windows Python3.13 wheel build has been removed because numpy < 2.0 does not support Python 3.13 but openfermion did not support numpy >= 2.0.
Now openfermion supports numpy >= 2.0 (ref: https://github.com/quantumlib/OpenFermion/issues/1008#issuecomment-2940961529 ), so we can support Python 3.13.


forest1040 (approved):
👍

kwkbtr:
Is it possible to publish Windows Python 3.13 binary to PyPI?

PR #689, opened 2025-07-02
Fix: fail cuda compilation in HPC SDK

aiueo-1234:
## Abstract
In HPC SDK, `FindCUDA` cannot find `curand` and `cublas`, so I changed it to `FindCUDAToolkit`.   
Also, since HIP language support is available in cmake version 3.21 or later, I changed `cmake_minimum_required` to 3.21.  

## Tested Environment
HPC SDK container: nvcr.io/nvidia/nvhpc:25.1-devel-cuda12.6-ubuntu24.04

cf)
https://cmake.org/cmake/help/latest/module/FindCUDA.html
https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html
https://rocm.docs.amd.com/en/latest/conceptual/cmake-packages.html

### DIFF
--- a/doc/en/source/intro/1_install.md
+++ b/doc/en/source/intro/1_install.md
@@ -46,7 +46,7 @@ Qulacs is tested on the following systems.
 - macOS Big Sur 11
 - Windows Server 2019
 
-If you encounter some troubles, see [troubleshooting](http://qulacs.org/md_4__trouble_shooting.html).
+If you encounter some troubles, see {doc}`2_faq`.
 
 ### How to install
 
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Qulacs is tested on the following systems.
 - macOS Big Sur 11
 - Windows Server 2019
 
-If you encounter some troubles, see [troubleshooting](http://qulacs.org/md_4__trouble_shooting.html).
+If you encounter some troubles, see [troubleshooting](https://docs.qulacs.org/en/latest/intro/2_faq.html).
 
 ### How to install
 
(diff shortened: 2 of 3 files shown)


KowerKoint (commented):
Please update `os` in `https://github.com/qulacs/qulacs/blob/main/doc/en/.readthedocs.yaml` to `ubuntu-24.04` so that cmake >= 3.21 is installed.
cf. https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-os

aiueo-1234:
> Could you also update cmake version requirement on https://github.com/qulacs/qulacs/blob/main/doc/en/source/intro/1_install.md and https://github.com/qulacs/qulacs/blob/main/doc/ja/source/intro/1_install.md (it also has old description about other requirements, so updating it to README's is better 🙇)?

OK. I'll do that.

doc/en/source/intro/1_install.md
```diff
@@ -27,24 +27,26 @@ A binary that is installed via pip command is optimized for Haswell architecture
...
+Note that `qulacs-gpu` includes CPU simulator. You don't need to install both.
 
 Qulacs is tested on the following systems.
 
 - Ubuntu 20.04
 - macOS Big Sur 11
 - Windows Server 2019
 
-If you encounter some troubles, see {doc}`2_faq`.
+If you encounter some troubles, see [troubleshooting](http://qulacs.org/md_4__trouble_shooting.html).
```
KowerKoint, line 49: No such file.

aiueo-1234:
It looks like workflows need to be allowed by the maintainer.  
https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks  
image

PR #700, opened 2026-01-23
update runner image of macos

KowerKoint:
runner `macos-13` was retired.
This PR update runner to use `macos-15`.


Glacialte (approved):
LGTM

PR #705, opened 2026-02-03
Code to add the ECR gate

usendon:
Hi,
I have written the code to add the ECR gate to Qulacs. I have already explained the changes I made in an issue that I sent, but I will repeat them here for clarity. Almost all my code is and addition to the original; I have only modified the following from the original code:
- MPIutil.cpp: mpitag ^= 1 << 19 (previously 20). This is necessary for it to work properly on my machine.
- pyproject.toml: license = { text = "MIT" } # previously license = "MIT"
    #license-files = ["LICENSE"] # previously uncommented. Because it gave me an error saying it was obsolete.

The only thing missing is the implementation of ECR with CUDA, which is what I am working on now. When I am done, I will make another pull request to complete it.

Uxía


usendon:
Thank you for the kind words! I am glad to contribute to Qulacs. 

I just finished removing all my comments, blank spaces and modifications to unrelated parts of the codebase. Now my PR should only contain the code necessary for the ECR gate.

Please let me know if you need anything else.

src/csim/update_ops_named_ECR.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+
+        __m128d a_lo = _mm_loadu_pd(ptr00); 
+        __m128d a_hi = _mm_loadu_pd(ptr01); 
+        __m128d b_lo = _mm_loadu_pd(ptr10); 
+        __m128d b_hi = _mm_loadu_pd(ptr11); 
+
+        auto mul_by_i = [](__m128d x) -> __m128d {
+            __m128d swapped = _mm_shuffle_pd(x, x, 0x1);
+            const __m128d sign = _mm_set_pd(1.0, -1.0);
```
KowerKoint: This should be _mm_set_pd(-1.0, 1.0), isn't it?
Suppose `x` shows $a+bi$, `swapped` is $b+ai$, and result will be $b-ai$ with this `sign`.

src/csim/update_ops_named_ECR.cpp
```diff
@@ -0,0 +1,455 @@
...
+        __m128d i_a_hi = mul_by_i(a_hi);
+        __m128d i_a_lo = mul_by_i(a_lo);
+
+        __m128d tmp_new_v00 = _mm_add_pd(a_hi, i_b_hi);
+        __m128d tmp_new_v01 = _mm_sub_pd(a_lo, i_b_lo);
+        __m128d tmp_new_v10 = _mm_add_pd(b_hi, i_a_hi);
+        __m128d tmp_new_v11 = _mm_sub_pd(b_lo, i_a_lo);
+
+
+        __m128d svec = _mm_set1_pd(sqrt2inv);
```
KowerKoint: As `_mm_set_1_pd` copies the argument to each 64-bit elements, `svec` will be $\frac{1}{\sqrt{2}}+\frac{1}{\sqrt{2}}i$, which is wrong.

src/csim/update_ops_named_ECR.cpp
```diff
@@ -0,0 +1,455 @@
...
+        state[basis_index_00] = new_v00;
+        state[basis_index_01] = new_v01;
+        state[basis_index_10] = new_v10;
+        state[basis_index_11] = new_v11;
+    }
+
+} 
+
+#ifdef _USE_SIMD
+void ECR_gate_parallel_simd(UINT target_qubit_index_0,
```
KowerKoint: Could you try using AVS-512 family intrinsics?
Other gate implementations (CX for handling two target qubits, Y for multiplying $i$) may help you.

src/csim/update_ops_named_ECR.cpp
```diff
@@ -0,0 +1,455 @@
...
+            svfloat64_t i_01 = mul_by_i(svptrue_b64(),  input01);
+
+
+            svfloat64_t output00 = svadd_x(svptrue_b64(), input01, i_11); 
+            svfloat64_t output01 = svsub_x(svptrue_b64(), input00, i_10); 
+            svfloat64_t output10 = svadd_x(svptrue_b64(), input11, i_01); 
+            svfloat64_t output11 = svsub_x(svptrue_b64(), input10, i_00); 
+
+
+            svfloat64_t sv_factor = svdup_f64(sqrt2inv);
```
KowerKoint: This also make $\frac{1}{\sqrt{2}}+\frac{1}{\sqrt{2}}i$.

src/csim/update_ops_named_ECR.cpp
```diff
@@ -0,0 +1,455 @@
...
+}
+
+#endif  // _USE_SVE
+
+
+
+
+#include <complex>
+#include <cstdio>
+#include <type_traits>
```
KowerKoint: Please remove this `include`, or move into `_USE_MPI` section if needed.

.gitignore
KowerKoint: Please revert this personal change.

KowerKoint (changes requested):
Thank you for responding to the request quickly.

I checked the changes except `gpusim/`. It seems that there are some problems on detailed implementations. Please check my comments.

If you do not really understand about our usage of AVX, SVE and MPI, it is OK to leave the implementation blank. You can fallback to `_unroll` functions, call general `multi_qubit_dense_matrix_gate` function, or  just throw `NotImplementedException`. We will implement these versions later.

In addicion, Could you put unit tests in `test/cppsim/test_gate.cpp`?

If you have some question, feel free to ask me.

usendon:
Thank you very much for your comments!
I prefer to implement the full version of the ECR gate. I will carefully review all your comments and make all the necessary changes to correct my implementation. 
It may take me some time so I will update the changes one by one.
Thank you again.
Uxía

src/csim/update_ops_named_ECR.cpp
```diff
@@ -0,0 +1,455 @@
...
+    const ITYPE amplitude_block_size = rtgt_blk_dim << 1; 
+
+#pragma omp parallel for
+    for (state_index = 0; state_index < dim;
+        state_index += amplitude_block_size) {
+        for (ITYPE offset = 0; offset < rtgt_blk_dim; ++offset) {
+           
+            const ITYPE idx0 = state_index + offset; 
+            const ITYPE idx1 = idx0 + rtgt_blk_dim; 
+            const std::complex<double> si0 = si[idx0];
```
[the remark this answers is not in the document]
usendon: Should I use CTYPE instead of CPPCTYPE, as defined in src/csim/type.hpp?

src/csim/update_ops_named_ECR.cpp
usendon: I revised all my code and I think that this is not a problem. If I understood correctly, `svfloat64_t sv_factor = svdup_f64(sqrt2inv);` constructs a vector in which all elements are $\frac{1}{\sqrt{2}}$. In output11 the elements are stored in the form (r1, i1, r2, i2, ...), where r* is the real part of a complex * and i* is the imaginary part. Multiplying output11 by sv_factor I obtain another vector like output11 but with all elements multiplied by $\frac{1}{\sqrt{2}}$. I am not adding any complex number i that modifies the result or its sign. 
I also checked my final statevectors with those generated by Qiskit and they are identical. 
Please tell me if I am missing something or if I didn't explain what I did with enough clarity.
Thank you!

src/csim/update_ops_named_ECR.cpp
usendon: I think it is the same as in SVE, so I believe I don't need to change this, it works correctly.

usendon:
Hello again! 
I just finished revising my code based on all your comments. I’ve marked all the comments I’ve addressed as resolved and modified my code accordingly. I also have a question and two responses to three of your comments, which I’ve written right below each one.
I think everything is working correctly now. When you have a chance, please review my code and let me know if anything else needs to be changed.
Thanks!

usendon:
Hi,
I just discovered an issue with my MPI implementation. Please wait until I make another commit to revise this function.
Thank you!

usendon:
Hi, everything is corrected now. When you have a chance, please review my code and let me know if anything else needs to be changed.
Thank you.

src/gpusim/update_ops_named.cu
```diff
@@ -354,6 +354,102 @@ __host__ void SWAP_gate_host(unsigned int target_qubit_index0,
...
+
+        some_v10 = make_gpuDoubleComplex(
+            -gpuCimag(state_gpu[basis01]),
+            gpuCreal(state_gpu[basis01]));
+
+        some_v11 = make_gpuDoubleComplex(
+            gpuCimag(state_gpu[basis00]),
+            -gpuCreal(state_gpu[basis00]));
+
+        tmp00 = cuCmul(make_cuDoubleComplex(sqrt2inv,0), cuCadd(state_gpu[basis01], some_v00));
```
KowerKoint, line 416: Please use `gpuC*` functions in stead of `cuC*`.
They are defined in `src/gpusim/gpu_wrapping.h`. These wrappers resolve the difference between CUDA and HIP.

src/cppsim/gate_named_two.hpp
KowerKoint, line 132: Please define `ECRGate` branch

src/csim/update_ops_named_ECR.cpp
```diff
@@ -0,0 +1,414 @@
...
+                    si = state + w*dim_work; 
+                    rw = w;
+                } else {
+                    ITYPE ideal_global_index_work = ((ITYPE)rank*num_work + w) * dim_work; 
+                    ITYPE my_target_index = ideal_global_index_work ^ ((1ULL << target_qubit_index_0) + (1ULL << target_qubit_index_1)); 
+                    rw = (ITYPE)(my_target_index%dim)/dim_work; 
+                    si = state + rw*dim_work; 
+                }
+            } else { 
+                si = state;
```
KowerKoint, line 327: Please initialize `rw` for this branch.

KowerKoint (changes requested):
I'm sorry for very late review.
There are some changes to do.

### Formatting

> This change does not pass the repository’s formatting checks. `clang-format-14 --dry-run --Werror` reports violations in several modified C++ files, particularly `update_ops_named_ECR.cpp`, and `git diff --check` reports multiple trailing
> whitespaces. Please run `./script/format.sh` and the configured Python formatters before merging.

### Python type stubs

> The runtime Python bindings expose `gate.ECR` and `QuantumCircuit.add_ECR_gate`, but the corresponding type stubs have not been updated. Please add `ECR` to both `pysrc/qulacs/gate.pyi` and `pysrc/qulacs/gate/__init__.pyi`, and add `add_ECR_gate` to
> `pysrc/qulacs/__init__.pyi`, so that type checkers and IDE completion recognize the new API.

PR #708, opened 2026-02-25
Fix notfound mpi.h

aiueo-1234:
When using a custom-installed Open MPI, the path to `mpi.h` is not appended to `CPLUS_INCLUDE_PATH`. Therefore, I changed `CMakeLists.txt` to use ` FindMPI`.

By the way, I noticed that the test command in CONTRIBUTING.md might be `make pythontest` instead of `make test`. Is that correct?


KowerKoint (approved):
LGTM

PR #712, opened 2026-03-03
apt update before installing libboost-dev / use local method for CUDA install

KowerKoint:
CI was failing for (temporal) apt condition.
This PR changes that:
- run `sudo apt update` before `sudo apt install libboost-dev`
- use local method for CUDA installing (same as https://github.com/qulacs/scaluq/pull/309 )


KowerKoint:
Thank you for your reviewing, but this branch includes the latest `main` branch. I think that `⚠️ Report is 2 commits behind head on main.` just shows that this branch has changes of two commits.

serowri (approved):
Thank you, LGTM!

PR #720, opened 2026-08-02
fix: don't let optimize_light skip over parametric gates when merging

### LINKED ISSUE #718: `optimize_light` incorrectly merges non-parametric gates across a parametric gate barrier, corrupting circuit semantics
`QuantumCircuitOptimizer::optimize_light` can silently produce a circuit with a different unitary than the original when a `ParametricQuantumCircuit` contains non-parametric gates on both sides of a parametric gate on the same qubit.

The optimized circuit runs without error and returns a well-formed, normalized state vector. The corruption is silent.

```python
from qulacs import ParametricQuantumCircuit, QuantumState
from qulacs.circuit import QuantumCircuitOptimizer as QCO
import numpy as np

circuit = ParametricQuantumCircuit(3)
circuit.add_X_gate(0)
circuit.add_parametric_RX_gate(0, 1.0)
circuit.add_Y_gate(0)

original = QuantumState(3)
circuit.update_quantum_state(original)

optimized_circuit = circuit.copy()
QCO().optimize_light(optimized_circuit)

opt = QuantumState(3)
optimized_circuit.update_quantum_state(opt)

print(original.get_vector())   # [ 0.-0.878j, -0.479+0.j, ... ]
print(opt.get_vector())        # [ 0.-0.878j, +0.479+0.j, ... ]  ← sign flipped
print(np.allclose(original.get_vector(), opt.get_vector()))  # False
```

## Root Cause

The algorithm in `optimize_light` tracks, for each qubit, the index of the most recently seen gate via `current_step`. When two consecutive same-qubit gates pass the `std::includes` subset check, they are candidates for merging. The code correctly rejects the merge when either gate is parametric but then continues without updating `current_step` for the current gate's qubits.

`src/cppsim/circuit_optimizer.cpp`, lines 204–207:

```cpp
// parametric gate cannot be merged
if (circuit->gate_list[pos]->is_parametric() ||
    gate->is_parametric())
    continue;          // ← current_step 
```

Because `current_step` is not updated, the parametric gate is invisible to the tracker. The next non-parametric gate on the same qubit looks up current_step, finds the entry from before the parametric gate, and merges with that stale, non-adjacent candidate, effectively reordering gates across a parametric gate it cannot commute past.

## Proposed fix

Update `current_step` before doing `continue`, so the parametric gate acts as an opaque barrier, exactly as the else branch already does for non-mergeable gates:

```cpp
if (circuit->gate_list[pos]->is_parametric() ||
    gate->is_parametric()) {
    for (auto target_qubit : target_qubits) {
        current_step[target_qubit] = make_pair(ind1, target_qubits);
    }
    continue;
}
```
This ensures no subsequent gate can claim the pre-parametric-gate entry as its merge partner, preserving the intended circuit ordering around parametric gates.

kriss-u:
## Summary

- Fixes a bug in `QuantumCircuitOptimizer::optimize_light` where a parametric gate sandwiched between two mergeable non-parametric gates on the same qubit could be silently skipped, causing the two non-parametric gates to merge directly and produce an incorrect circuit/unitary.

- Adds a regression test, verified to fail against the pre-fix code and pass with the fix.

Fixes #718


serowri (approved):
Looks good to me,  It works correctly. 
Thank you for contributing!

kriss-u:
@serowri The checks have passed. I think, this is ready to merge.

serowri:
Thank you for contributing!

